PR #20, opened 2015-12-08
pull master changes

ebroecker:
@altendky: I think, you are right. The 'ints' only force python to handle the variables as numbers (instead of strings)

By the way: this pull-request is from master to byteorder-branch...


altendky:
* Is it safe to assume they are all `int()`s?  It seems like the factor could be a float and at least on export also the max.
* Should the same be done with the minimum?
* I think this would affect some of the test results so they should be included.

PR #25, opened 2016-01-07
(Optionally) enforce output format using command line flag.

m8ddin:
This change adds a command line flag (-f) to enforce the output format,
ignoring the output file extension.
This is eases the integration of canconvert into a diff tool (Beyond Compare).

Would be nice, if the functionality would be integrated into the main branch.
Thanks in advance!
Martin

Flag usage: -f <file extension> (with or without leading '.')
	-f csv
	 f .csv
	-f kcd
	etc.


ebroecker:
Thanks for the helpful patches which make the tool better!

PR #34, opened 2016-02-05
Added all information required for decoding in the exported json file

reinzor:
Not every information required for parsing was present in the exported json file; also changed the signals into a list; not a dictionary with start bit string keys.

### DIFF
--- a/canmatrix/exportJson.py
+++ b/canmatrix/exportJson.py
@@ -43,10 +43,18 @@ def exportJson(db, filename):
     exportArray = []
 
     for frame in db._fl._list:
-        signals = {}
+        signals = []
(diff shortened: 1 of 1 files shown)


reinzor:
Thanks!

-Rein

ebroecker:
seems, that the json is not compatible to https://github.com/ericevenchick/CANard anymore? 
I think this doesn't matter, because CANard seems to be dead anyway...

one more information is missing: Extended-Flag for the frames.
Do you have any preferences for this flag (any name, any spec for your json format)?

reinzor:
I think a boolean 'is_extended_frame' would do.

-Rein

PR #41, opened 2016-04-29
Split CAN Id information to join frame signals

swatz:
Join signals bases on the startbit if they have the same PGN
(ps : this is my first pull )

modify : canmatrix.py
add  examples with files


altendky:
I would suggest that j1939 specific stuff be flagged as such. I didn't look over it much but it might make sense to have a separate j1939 file?  Maybe not, just a thought.

swatz:
In the future I was thinking of commands like
> canjoin.py --j1939 --on='signal' --by='startbit length' source1.dbc source2.dbc target.dbc
or
> canjoin.py --j1939 --how='inner' --on='pgn sa' --by='startbit' source1.dbc source2.dbc target.dbc

So will create a separate file

ebroecker:
I don't know much about j1939, maybe I take this pull request as trigger for learning something about...

As far as I see there is no conflict with the current code.
Maybe flagging with j1969 could be an option.

My suggest:
I merge this code to mainline because it does not interfere with the current code and may help somebody to solve his issue with our code. This is always good.

If we get more j1939-specific code we could refactor the code later.

I'm even not sure at the moment, if a new tool "canjoin.py" is needed, or if it is integrateable in convert.
My intention till now was to create a "swiss army knife" named (can)convert.py (and maybe sometimes a GUI for it also). But as I have nearly no knowledge about j1939 I could be wrong with my wishes here...     

please shout - if anybody thinks merging does hurt ?!

Ps: Nobody has to excuse himself for doing his first pull. JUST DO IT!
This project is my first github project. I'm also not very familiar with the correct workflow and so on.

ebroecker:
@swatz

Do you work in "canjoin" these days?

I'd like to rename CanId to j1939CanId and join.py to j1939join.py 

Is this OK for you?
Any suggestions?

PR #44, opened 2016-05-25
export_fix: insert "" for empty string (exportDbc)

floGik:
Exported DBC files with empty signal attributes failed on reimport. This is fixed by inserting quotation marks (`""`) for empty value strings.

(Error was: `NoneType object has no attribute group` because the regex doesn't match without the quotation marks, which is expected behaviour)

Also changed checking for empty `dict` by using [Python's ability to evaluate empty lists and dicts to `False`](https://docs.python.org/3.5/library/stdtypes.html#truth-value-testing). Feel free to reject this cosmetic alteration.


ebroecker:
thanks

PR #48, opened 2016-07-08
Conversion fixes, kcd, unify/fix startbit get/set

dmahurin:
Correct kcd import/export.
Correct set/getMsbStartbit set/getMsbReverseStartbit functionality.
Unify all startbit set/get functions.
Use internal startbit with consistent bit numbering (also simplifies bit math).


ebroecker:
wow - seems you have invested a lot of work in this code.
Tanks for cleanup this boring byteorder-issue.
Seems you have a lot of knowhow about the differrent formats. Maybe you could document some info about the different formats in the wiki - just if you find time.

Thanks again for this big pullrequest!
Eduard

PR #59, opened 2016-08-23
Add exportany function

christiansandberg:
The `exportany` function should basically be the counter-part to `importany`.

Also fixed an issue with xls export from my last pull request.


ebroecker:
importany returns a dictionary with possible multiple matrix-object.
exportany currently only takes a single matrix for export. 
This should be changed once. 
Maybe I'll find the time to do this.

Anyway I merged the pull ...

PR #72, opened 2016-12-22
Fix token parsing for BA_ section

LucasBuckland:
Previously there was an issue with token parsing in the _BA section of DBC files. If the parser ran into a '[' or ']' within quotes then it would cough. This commit fixes this bug and allows the parser to work for these cases.


ebroecker:
I already stole your patch and merged it in dev-0.6-branch

LucasBuckland:
Should I be using the dev-0.6-branch instead of the master branch then? Is the code more mature there?

PR #74, opened 2017-01-16
Changes for dependency handling

christiansandberg:
Reduce logging level for unsupported formats at import and signal error on load/dump instead. Some formats may not need to be supported if never used.
YAML format only depends on [PyYAML](https://pypi.python.org/pypi/PyYAML), not [pyaml](https://pypi.python.org/pypi/pyaml).
Always depend on future since it can be installed on Python 3 as well without any conflicts.

### DIFF
--- a/canmatrix/formats.py
+++ b/canmatrix/formats.py
@@ -17,7 +17,7 @@ for module in moduleList:
         import_module("canmatrix." + module)
         loadedFormats.append(module)
     except ImportError:
-        logger.error("Error importing canmatrix." + module)
+        logger.info("%s is not supported", module)
 
 for loadedModule in loadedFormats:
     supportedFormats[loadedModule] = []
@@ -53,6 +53,7 @@ def loadp(path, importType=None, key="", flatImport=None, **options):
     if importType:
         return load(fileObject, importType, **options)
     else:
+        logger.error("This file format is not supported for reading")
         return None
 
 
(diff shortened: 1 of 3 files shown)


altendky:
We are using `__future__` not `future` so I think the dependency on `future` is wholly unneeded.  (and I think a blame would point at me, sorry...)

christiansandberg:
We're actually using the `builtins` module which exists natively in Python 3 but in Python 2 is implemented by the future package. So it is not obvious that it is a dependency.

altendky:
Mm, so I wasn't crazy before.  How about something like below.  But, I honestly haven't internalized what dependencies should be specified where (setup.py vs. requirements.txt).  Also, this is untested.

```
extras_require={
    ':python_version<"3.0"': ['future']
}
```

http://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies
https://www.python.org/dev/peps/pep-0426/
 
Hey!  Just noticed you're the CANopen library author.  I don't expect to use it much in my current job but I've done a good bit of CANopen work before and was excited when a friend pointed out your library on PyPi.

christiansandberg:
Tried that in 756c47c but it fails in Travis, probably due to older pip or wheel versions. :(

Yeah I thought that there probably are a lot of proprietary CANopen implementations but couldn't find one that was open and not specific to some vendor so I open sourced my library and hoped that it could be useful to others as well.

PR #98, opened 2017-10-03
Adding encode and decoding methods to Signals, Frames and CanMatrix

pierreluctg:
Adding methods that allows using canmatrix for generation and parsing of can frames

### DIFF
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -175,10 +175,13 @@ def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
 
 
-class SignalValue(object):
+class SignalValue(float):
     """Proxy class to hold a Signal and value.
     Allows to have correct string representation of the signal value
     """
+    def __new__(cls, signal, value):
+        return float.__new__(cls, value)
+
     def __init__(self, signal, value):
         self._signal = signal
         self.value = value
@@ -191,16 +194,6 @@ class SignalValue(object):
             return self._signal.values.get(self.value)
         elif self._signal.unit and self._signal.unit not in ['SED', 'Mixed']:
             return str(self.value) + ' ' + self._signal.unit
-
-
-class SignalValueInt(SignalValue, int):
-    def __new__(cls, signal, value):
-        return int.__new__(cls, value)
-
-
-class SignalValueFloat(SignalValue, float):
-    def __new__(cls, signal, value):
-        return float.__new__(cls, value)
 
 
 class Signal(object):
@@ -478,7 +471,10 @@ class Signal(object):
         :return: str
         """
         endian = '<' if self._is_little_endian else '>'
-        bit_type = 's' if self._is_signed else 'u'
+        if self.is_float:
+            bit_type = 'f'
+        else:
+            bit_type = 's' if self._is_signed else 'u'
         return endian + bit_type + str(self._signalsize)
 
     def encode(self, value=None):
@@ -499,9 +495,14 @@ class Signal(object):
                 raise ValueError(
                     "{} is invalid value choice for {}".format(value, self)
                 )
-        if self._min > value > self._max:
-            raise ValueError()
-        physical_value = int((int(value) - self.offset) / self.factor)
+        if not (self._min <= value <= self._max):
+            raise ValueError(
+                "Value {} is not valid for {}. Min={} and Max={}".format(
+                    value, self, self._min, self._max)
+            )
+        physical_value = (value - self.offset) / self.factor
+        if not self.is_float:
+            physical_value = int(physical_value)
         return physical_value
 
     def decode(self, value):
@@ -511,9 +512,7 @@ class Signal(object):
         :return: SignalValue
         """
         value = value * self.factor + self.offset
-        if self.is_float:
-            return SignalValueFloat(self, value)
-        return SignalValueInt(self, value)
+        return SignalValue(self, value)
 
     def __str__(self):
         return self._name
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -31,7 +31,7 @@ from __future__ import division
 
 import logging
 import math
-from collections import namedtuple
+from collections import OrderedDict
 
 import bitstruct
 from six import string_types
@@ -173,6 +173,31 @@ class BoardUnitList(object):
 
 def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
+
+
+class SignalValue:
+    """Proxy class to hold a Signal and value
+
+    Not using class "SignalValue(object):" on purpose to proxy magic methods
+    (__eq__, __lt__, etc.) and other methods from value or signal
+    """
+    def __init__(self, signal, value):
+        self._signal = signal
+        self._value = value
+
+    def __getattr__(self, item):
+        if hasattr(self._value, item):
+            return getattr(self._value, item)
+        return getattr(self._signal, item)
+
+    def __str__(self):
+        return self._get_value_string() or str(self._value)
+
+    def _get_value_string(self):
+        if self._signal.values and self._value in self._signal.values:
+            return self._signal.values.get(self._value)
+        elif self._signal.unit and self._signal.unit not in ['SED', 'Mixed']:
+            return str(self._value) + ' ' + self._signal.unit
 
 
 class Signal(object):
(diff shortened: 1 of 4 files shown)


ebroecker:
@pierreluctg:
Thanks for this code.
This feature has often been asked for.

From my point of view there is nothing against merging.
@altendky : you have some open questions, so should we wait with merging?

I'll look for some test-data. 
There is some in https://github.com/reinzor/libcan-encode-decode (which does the same but in C)...

ebroecker:
sorry - didn't want to close ...

pierreluctg:
Looks like there is an issue with new change on Python 3.x. working on a fix

canmatrix/canmatrix.py
```diff
@@ -170,6 +175,31 @@ def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
 
 
+class SignalValue:
+    """Proxy class to hold a Signal and value
+
+    Not using class "SignalValue(object):" on purpose to proxy magic methods
```
altendky: Are you saying you can't inherit from `object` (in other words, you have to use old style classes in py2) to specify your own `__getattr__`?  If so, [my quick check](https://repl.it/MExN) seems to work fine.

altendky:
I've been using pytest in my own stuff.  It seems reasonable to settle on that if we were to switch to a 'real' testing framework rather than whatever I stuck in there before adding py3 support.

@ebroecker That's up to you, of course.  I would think it would be worth a little consideration of both the couple details I brought up and how it would be good to add this on.  But, this wouldn't break anything I have so it's not really a big deal.  As I mentioned, my code only uses canmatrix for parsing, basically.  Probably the most concrete thing to consider is using either `future` or `six` for 2/3 compatibility, but not both.

Presently, the encoding and decoding here are located on the signal and frame objects but if the latest received values are to be retained that is still an external activity.  Perhaps the signals and frames should have value, human value, and bytes sorts of properties to hold onto this?  Make a wrapper function for the decode that assigns back to the signals?  I see that the `SignalValue` class was added which creates some relationship but only one way and I'm not sure that hybridizing `Signal` with `int` or `float` etc really results in a sensible object.  I might be missing it but how do I get the value from a `SignalValue`?  `my_signal_value._value`?

The encode and decode probably need an option to ignore the min/max limits and a helper-function to just check that perhaps after decoding.  It should probably be a custom canmatrix exception class as well rather than ValueError so it can be caught independently.

I don't think I'm arguing against this code or merging, just trying to think things through beforehand.

pierreluctg:
@altendky, thank you for the comments. 

You are correct, `SignalValue` is there to hold the actual value of a signal and allow proper `str` representation. I wanted to keep the signal definition `Signal` separated from the value.

Let say you hare a CAN sniffer that reads frames from the bus and prints the decoded data. Once you have parsed you DB, the `Signal` will be static. You only get various different `SignalValue`.

From a `Signal` class instance, you do not need access to `SignalValue`.

To get the value from a `SignalValue` you do not have to access `_value` or `value`, its a like a proxy class. I made few modification in the last commit to make it easier. So `my_signal_value` behave just like a int or a float (depends on signal.is_float).

pierreluctg:
I have removed `six` now only using `future`

canmatrix/canmatrix.py
pierreluctg: `__getattr__` works bu not for __eq__, __lt__, etc.

altendky: I was thinking on my lunch walk and agree that my example is irrelevant.  I'll try again.  But, if this truly only can be done on old-style classes how would it work in Python 3?

altendky:
@pierreluctg Sure, if you only want the value at the time of decoding that's fine and really requires no class at all.  But, what if you want a place to lookup the most recent value at any given point in time?  I guess I just don't see much need for the pass-through class since we could instead just use regular methods for finding a signal object and then looking up the value in a dict keyed by the signal.  Or just keep it on the `Signal` itself.

Looking at the new approach, messing with `__new__` seems pretty questionable.  This seems like a lot to avoid adding one property and one method to Signal.  If separate, probably just have a `.signal` attribute on the new class to get at the signal when you want it.

canmatrix/canmatrix.py
altendky: Alrighty, you shifted away from that, but, FYI, perhaps [`__getattribute__()`](https://docs.python.org/2/reference/datamodel.html#object.__getattribute__) was what you were missing?

tests/test_codec.py
```diff
@@ -0,0 +1,170 @@
...
+                'Signal': u'two'
+            }
+            data_bytes = tuple(bytearray(bus.encode(test_frame1, data)))
+
+            decoded = bus.decode(test_frame1, data_bytes)
+
+            for k, v in data.items():
+                assert str(decoded[k]) == v
+
+    def test_encode_decode_signal_value_choice_str(self):
```
altendky: In py2, `str` is a byte sequence and, basically, should not be used.  It does wacky automatic stuff on top of just being confusing because of it's misleading name.

We probably want a single function and:
```
from __future__ import unicode_literals
from builtins import *
```
http://python-future.org/compatible_idioms.html#unicode-text-string-literals
http://python-future.org/compatible_idioms.html#basestring

If someone wants to encode a string to bytes, they need to pick their encoding which we can't do for them usefully.

I don't generally like the `*` in `from builtins import *` but I sort of feel like this is a special case where we want to make certain that as much as possible is made to be common.

pierreluctg: This is not encoding a string to bytes. `'two'` is a value description for the signal `Signal` in the frame `0x123`.

The value getting encoded will be `2`

canmatrix/canmatrix.py
pierreluctg: Same issue with `__getattribute__` see the note in documentation.

tests/test_codec.py
altendky: Yeah, pretty sure I misread this too...  again, I'm sorry.

Still, I think the imports are relevant for consistency of the environment the code runs in.  `str` should be a Unicode character sequence in keeping with elsewhere.  I may have messed up not importing `unicode_literals` elsewhere, I'm not sure.

Also, I don't think we want to be testing to make sure that people can pass byte sequences (py2 `str`) where we expect them to pass a Unicode sequence.  Sure, py2 does goofy stuff but it shouldn't work in py3.  Presently I'm thinking this would pass in py2 because of the automatic bytes->unicode conversion and in py3 because of `''` and `u''` being the same thing anyways.  With the imports I would expect both to pass in both py2 and py3 because they are identical at that point.  If you changed one to be `b''` then it would fail in both py2 and py3.  If I'm predicting correctly...

pierreluctg: I fully get your point now. SO you suggest removing `test_encode_decode_signal_value_choice_str`?

altendky: Along with adding the imports, sure.

canmatrix/formats.py
```diff
@@ -43,18 +43,18 @@ def loads(string, importType=None, key="", flatImport=None, **options):
...
-            if path.endswith(extension) and "load" in supportedFormats[key]:
-                importType = key
-                break
-
-    if importType:
-        return load(fileObject, importType, **options)
-    else:
-        logger.error("This file format is not supported for reading")
-        return None
+    with open(path, "rb") as fileObject:
```
altendky: Ack!  Yeah, much better...

altendky:
I'm still a bit skeptical of `SignalValue`.  Messing with `__new__`, making a partially mutable `float`, etc.  The benefit just seems very slight for the funny business.

I believe the benefits in the present mechanism (`SignalValue` and the `OrderedDict` with signal name keys) are:

1. Decoded message can be indexed by the signal's name directly
2. Signal is accessible as well
3. Value is accessible
4. Automatically converts to a string

Perhaps we can still shift some functionality to a custom container such as a `DecodedFrame`?  Perhaps maintain a similar interface to the existing frames with `.signalByName()` and `.valueByName()` and `.strByName()` for quick by-name access.  One of those could be used in `__getitem__()`, though that might be better as a whole lib feature addition if it were to happen.  Or shift `.strByName()` onto a regular value class that just has the raw, human, and str attributes (or sure, `__str__()` `:]`).  If you want the raw and human values to be modifiable you could make them properties and recalculate the other based on an attribute referring to the original signal or cached copies of the factor/offset.

In my case I decided I didn't want my entire application to be tightly coupled to canmatrix so I made my own duplicative structure of frames and signals etc.  But, if this feature is going into canmatrix itself, why not just add the value and string conversion to the regular `Frame` and `Signal` classes?

pierreluctg:
@altendky I agree that SignalValue is a bit messy. Will fix it.

PR #127, opened 2018-01-06
`Signal.is_multiplexer` is a calculated value

grissiom:
`Signal.is_multiplexer` is calculated from the `multiplex` parameter.
So it shouldn't set from the parameters.


ebroecker:
thanks

Issue #126, opened 2017-12-29
Scaling of min/max/default in signals


altendky:
Are the in-memory min/max/defaults supposed to be comparable to the raw integers in a CAN message?  or the human value they represent?  With a factor of 0.1 and a default human value of 15.1, should signal.attributes['GenSigStartValue'] be `'15.1'` or `'151'`?

This is coming up because some of my defaults are not getting written to my .sym file because after the scaling (introduced at the beginning https://github.com/ebroecker/canmatrix/commit/95611b1a8e4ebf22f14bd0ff08e348f66654c048#diff-2fda6f9ac5ab0214cac52642f17fc1f7R167) the value is not in the range of the min and max and so isn't output.  I just thought it would be good to know the intended in-memory value before trying to review the code.

For reference, here is a sample signal from the symbol editor.  Note that the max and default both have decimals so they are scaled values, not directly comparable to any raw integer value from a message.

```
Var=Var1 unsigned 0,8 /f:0.1 /min:11 /max:25.4 /d:15.1
```

ebroecker:
I know some dbc's where the default is value for a signal is 'invalid value' which is often not in range between min and max.
So maybe the min/max-check for the default-value is arguable.

In my opinion the raw-default-value should be used, because scaling often leads to some arithmetical errors. 

I think .dbc stores the raw-default value but .sym stores the scaled value?

altendky:
In addition to calculation errors, even just converting from a string to a float can induce error.  That is why I added the float factory option and use [decimal](https://docs.python.org/3/library/decimal.html) (https://github.com/ebroecker/canmatrix/commit/a622e5ebf2aac1fa226aea89474a24df6224c5b9).

But, not calculating is only an option if all file formats store all values either scaled or not.  Well, at least given the assumption that we want a consistent state in memory.  Alternatively we could calculate losslessly using decimal and/or [fractions](https://docs.python.org/3/library/fractions.html) as needed.  If some formats store scaled values and others raw/unscaled values then we are stuck doing calculations since the in-memory value should be consistent.

ebroecker:
... slowly but surely I understand this issue :-)
Well in my opinion fractions seem to be the best solution.

ebroecker:
A long time later, I am working on this issue.
I now decided to use the scaled values and store them  [decimal](https://docs.python.org/3/library/decimal.html) 

I have done this now for
```python
Signal.initial_value
```

PR #128, opened 2018-01-09
Fix float arithmetic issues when converting phys and raw values

pierreluctg:


altendky:
Why the conversions to `str` before `decimal.Decimal`?

pierreluctg:
@altendky this is in order to convert the string representation to Decimal and not Float to decimal.

```python
from decimal import Decimal

d = 0.2

print(Decimal(d))
print(Decimal(str(d)))
```

This will output
```python
0.200000000000000011102230246251565404236316680908203125
0.2
```

The best solution would be to create a Decimal object from the string stored in the DBC (or other format) directly. But I did no want to break other things.

altendky:
The problem is that Python can't tell the difference between which one was intended. Anyways, I added a `float_factory` option for the .sym import and this also came up in #126.  I would tend to offer either an option for all formats or just change the default. Though I would note that `fractions` might be better than `decimal` if we want to offer lossless calculations. Or just leave it to the user to pick what numeric strings become.

pierreluctg:
@altendky Thank you, I changed it to use `float_factory`

canmatrix/dbc.py
```diff
@@ -409,6 +410,8 @@ def load(f, **options):
     else:
         dbcCommentEncoding = dbcImportEncoding
 
+    float_factory = options.get('float_factory', decimal.Decimal)
```
altendky: Perhaps we should default to `decimal.Decimal` but doing it here would override the default of `float` in the `Signal` constructor.  If we like this `float_factory` interface and are going to spread it around, it should probably be consistent.

https://github.com/ebroecker/canmatrix/blob/a838e833befd666b32703bcf41fd2f0fa575c7a1/canmatrix/canmatrix.py#L201

canmatrix/dbc.py
```diff
@@ -475,18 +478,24 @@ class FollowUps(object):
...
-                                     factor=temp.group(6),
-                                     offset=temp.group(7),
-                                     min=temp.group(8),
-                                     max=temp.group(9),
-                                     unit=temp_raw.group(10).decode(
-                                         dbcImportEncoding),
-                                     receiver=receiver)
+                    extras = {}
+                    if float_factory is not None:
+                        extras['float_factory'] = float_factory
```
altendky: Without the above change to default `float_factory` to `None`, I think `float_factory` would always be `not None`.

canmatrix/canmatrix.py
```diff
@@ -33,13 +33,16 @@
...
+
 logger = logging.getLogger('root')
 try:
     import bitstruct
 except:
     bitstruct = None
     logger.info("bitstruct could not be imported // No signal de/encoding possible // try pip install bitstruct")
-#from past.builtins import basestring
+
+from past.builtins import basestring
```
altendky: I would rather not allow mixing of byte-strings (py2 `str`) and unicode strings.  Do we really need to?  I do see that the use of basestring was actually added elsewhere.

http://python-future.org/compatible_idioms.html#basestring

I'm fine with this being considered/addressed separately from this PR.

requirements.txt
```diff
@@ -3,3 +3,4 @@ xlwt
 xlrd
 xlsxwriter
 pyyaml
+future
```
altendky: Hmm...  Yes, this should be added, but it points out a separate issue.  I thought `requirements.txt` was mostly for specifying specific versions.  Such as using `pip freeze > requirements.txt`.  Though I'll admit that in my other projects I only enter direct dependencies and their versions into `requirements.txt`.  :[  I should stop doing that.

canmatrix/canmatrix.py
```diff
@@ -404,19 +408,15 @@ def phys2raw(self, value=None):
...
 
-    def raw2phys(self, value, decodeToStr = False):
+    def raw2phys(self, value, decodeToStr=False):
         """Decode the given raw value (= as is on CAN)
         :param value: raw value
         :return: physical value (scaled)
         """
 
-        value = value * self.factor + self.offset
+        value = self.float_factory(value) * self.factor + self.offset
```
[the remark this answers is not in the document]
pierreluctg: The thing is that the caller may not know what is the `float_factory` class.

requirements.txt
pierreluctg: Agree, I do not think this should be solve in this PR :)

canmatrix/canmatrix.py
pierreluctg: It is already in use in the code on line 396

pierreluctg:
@altendky I think all your comments have been addressed.

canmatrix/canmatrix.py
altendky: Right, that's why I don't intend to say it shouldn't be part of this change.  But, I am pointing out that I think it shouldn't be used at all.  @ebroecker, looks like you committed this.  Do we really want to allow py2 byte-strings to be considered strings?

https://github.com/ebroecker/canmatrix/commit/f949335a59e3bca70133e7809ea95a38d546dec3#diff-6e14cb4889f3ecdbe6c054170655f1b4R470

canmatrix/canmatrix.py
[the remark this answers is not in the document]
altendky: How?  They are writing the program that is calling this.  If they want a `decimal.Decimal` to be used, they should pass it in instead of a `float`, for example.  If they pass a `float` then converting it to a `decimal.Decimal` isn't going to do a lot of good as the loss of accuracy has already occurred.

But I see you changed it so no hold up here.

canmatrix/canmatrix.py
```diff
@@ -404,13 +408,9 @@ def phys2raw(self, value=None):
                 "Value {} is not valid for {}. Min={} and Max={}".format(
                     value, self, self.min, self.max)
                 )
-        raw_value = (value - self.offset) / self.factor
-
-        if not self.is_float:
-            raw_value = int(raw_value)
-        return raw_value
+        return (value - self.offset) / self.factor
```
altendky: If this is supposed to return a value for use on the CAN bus and it's not a floating point signal then the result should be an integer, yes?  Though there could be an argument about what conversion mechanism should be used, truncation vs round vs all the different rounding techniques.  But it should be an integer?

Perhaps there should be two functions.  One for scaling/offset and another that creates an actually bus-usable value?

Side note since this is pre-existing, `phys2raw` returns the physical value per the comment.  Seems backwards.  I would expect 'physical' to be in regular SI/imperial units and 'raw' to be bus values.

canmatrix/canmatrix.py
```diff
@@ -198,17 +202,17 @@ def multiplex(value):
                 multiplex = value
             return multiplex
 
-        float_factory = kwargs.pop('float_factory', float)
+        self.float_factory = kwargs.pop('float_factory', float)
```
altendky: As a nitpick, I think the `self` part could be dropped to tidy the diff.

canmatrix/canmatrix.py
ebroecker: not really - I even don't know anymore, what I wanted to check, if given 'value' is a string or a numeric value. I think it would be even more easy to check that given 'value' is numeric...

canmatrix/canmatrix.py
ebroecker: @altendky 
in dbc factor and offest mean:
`pyhiscalValue = valueInCanFrame * factor + offset;`

`phys2raw` should convert physicalValue to CanValue:
`valueInCanFrame = (pyhiscalValue - offset) / factor` looks right for me?

Is this another 'dbc-default' which should be documented somewhere?

altendky: @ebroecker

1) the doc string and function names seemed in conflict.

2) this change removes the conversion to an integer.  Unless the signal is floating point on the bus, shouldn't it be made an integer?  Unless we want to switch to having the concept of a bus-scaled value (that is still float, or decimal, etc) separate from an actually bus representable integer.

canmatrix/canmatrix.py
altendky: Or just have a separate function that converts a string? Then you don't have to do any type checking in either. Both either work or don't.

canmatrix/canmatrix.py
ebroecker: 1. I had to read once more...
the function name is right, BUT the doc-str is backward...

2. you are right

altendky: @pierreluctg

So, all clear on this now I think. You are intending to have this function return a bus-scaled value which is not actually usable on the bus because it still requires an int conversion?  (Truncation, rounding, etc). I don't have any use case for that in mind, but I haven't thought about it much. Generally, it seems like it might be sensible to have though. I'd probably have a different function name though.

canmatrix/canmatrix.py
pierreluctg: fixed

canmatrix/canmatrix.py
pierreluctg: fixed

altendky:
@pierreluctg tidying formatting is good but it might be better to handle that separately in the future.  The final diff here is mostly unrelated to the actual topic of the PR.

Issue #133, opened 2018-01-29
Unable to install `No module named past.builtins`


altendky:
I merged 7d8aa691e1e97af710e2aa81c4ca36748c83bfb1, so shame on me...  and then I missed the broken build too I guess.

1) We are using `basestring`...
 * As mentioned before, can we just not check types?  If we want the user to be able to pass two types, give them a separate function for each sort of different thing?
 * If it is kept and still needed for install we would need to set `future` as a `setup_requires` along with the existing `install_requires` reference.  Hopefully we don't need this though.
2) We are revisioning the version and it is inside the package too.
 * We are importing from our package in our `setup.py` to get `version`.  Versioning could be handled by Versioneer.
 * In our `__init__.py` we import stuff.  This means that nothing can be imported without importing all of those things.  This is questionable for developers using the lib but also  in general in this case since trying to import our version module triggers importing a significant chunk of canmatrix.

I'll start a branch for these things and see what opinions are had.

ebroecker:
I close this old bug, api has changed yet

PR #135, opened 2018-01-29
Create Signal.enum2raw() instead of using basestring in Signal.phys2raw()

altendky:
Issue #133 

@pierreluctg, I don't seem to be able to add you as a reviewer but please do weigh in on this if you want.

### DIFF
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -43,8 +43,6 @@ except:
     bitstruct = None
     logger.info("bitstruct could not be imported // No signal de/encoding possible // try pip install bitstruct")
 
-from past.builtins import basestring
-
 
 class FrameList(object):
     """
@@ -411,6 +409,23 @@ class Signal(object):
 
         return endian + bit_type + str(self.signalsize)
 
+    def enum2raw(self, name):
+        """Return the raw value (= as is on CAN)
+
+        :param name: enumerator name
+        :return: raw unscaled value as it appears on the bus
+        """
+        for value_key, value_string in self.values.items():
+            if value_string == name:
+                value = value_key
+                break
+        else:
+            raise ValueError(
+                "{} is invalid value choice for {}".format(name, self)
+            )
+
+        return self.phys2raw(value=value)
+
     def phys2raw(self, value=None):
         """Return the raw value (= as is on CAN)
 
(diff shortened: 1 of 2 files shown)


pierreluctg:
With this change, `raw2phys` (with `decodeToStr=True`) and `phys2raw` are no longer symmetrical. Do we want to keep them symmetrical?

Also, do we want to rename `decodeToStr` param to something else like `decodeEnum`?

altendky:
Good point.  I'd probably just make a `raw2enum()`.  Then there is no need for a parameter and the return type for each function is consistent.  I'll go update the PR in a couple minutes.

Enums should probably be real Python enumerations...  but that's a bit bigger thing to consider.  Even just having them centralized on a database rather than stored in each signal would be good.  I did some sort of hack on for that, maybe just for .sym, but I don't recall the details off the top of my head.  It was important because outputting to a .sym you need the enumeration definition once at the top of the file and then each signal only refers to the name from the definition.

pierreluctg:
@altendky, `Frame` encode and decode also need to be updated while keeping the ability to encode using values from enums and decode enums for signals with enums.

altendky:
@pierreluctg @ebroecker I think part of the issue here is that we are using strings at all.  Maybe I can make an interrim change to address the `Frame.encode()` and `decode()` methods but what do you think of having `enum.IntEnum`'s for enumerations rather than strings and integers?  This would force the conversion to/from an integer/enumerator to be done by the application code.  I guess that would mean that the interrim here would be to simply not support strings at all (remove the new `enum2raw()` and `raw2enum()` functions).

Here's a quick example of `enum.IntEnum`.

py2: https://repl.it/@altendky/basic-IntEnum-py2
py3: https://repl.it/@altendky/basic-IntEnum-py3

ebroecker:
@altendky 
is there anything todo or can I merge (after solving conflicts?)

altendky:
What thoughts do you have on my last comment?  We really need to get away from accepting two different types in a single parameter. It makes things much more confusing when you try to mix functionality.

ebroecker:
yes, you are surely right. But is this not exactly what this commit does ? 
There may be more places in the code where curently more types are allowed, but this one would be finished after this pr? 
Do I miss something?

altendky:
As I recall, this is incomplete as per @pierreluctg's last comment.  But I'll have to take a look next week.

ebroecker:
OK, thanks

PR #163, opened 2018-07-02
Attr PR - for discussing the changes

ebroecker:
This PR is only for discussing the changes.

### DIFF
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -240,23 +240,22 @@ class Signal(object):
 
     name = attr.ib(default = "")
     float_factory = attr.ib(default=float)
+    float_factory = float
     startBit = attr.ib(type=int, default=0)
     signalSize = attr.ib(type=int, default = 0)
     is_little_endian = attr.ib(type=bool, default = True)
     is_signed = attr.ib(type=bool, default = True)
-#    float_factory = attr.ib(default = float)
-    float_factory = float
-    offset = attr.ib(convert = float_factory, default = 0.0)
-    factor = attr.ib(convert = float_factory, default = 0.0)
+    offset = attr.ib(converter = float_factory, default = 0.0)
+    factor = attr.ib(converter = float_factory, default = 0.0)
 
     #    offset = attr.ib(converter = float_factory, default = 0.0)
 
-    min  = attr.ib(convert=float_factory)
+    min  = attr.ib(converter=float_factory)
     @min.default
     def setDefaultMin(self):
         return  self.calcMin()
 
-    max =  attr.ib(convert = float_factory)
+    max =  attr.ib(converter = float_factory)
     @max.default
     def setDefaultMax(self):
         return  self.calcMax()
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -29,6 +29,7 @@
 
 from __future__ import division
 import math
+import attr
 
 from collections import OrderedDict
 
@@ -185,7 +186,7 @@ class BoardUnitList(object):
         """
         remove Boardunit/ECU to list
         """
-        if BU.name.strip() not in self._list:
+        if BU.name.strip() in self._list:
             self._list.remove(BU)
 
     def byName(self, name):
@@ -221,6 +222,7 @@ def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
 
 
+@attr.s
 class Signal(object):
     """
     contains on Signal of canmatrix-object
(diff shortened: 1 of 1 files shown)


canmatrix/canmatrix.py
```diff
@@ -36,6 +36,13 @@
 import logging
 import fnmatch
 
+defaultFloatFactory = float
+try:
+    from decimal import *
+    defaultFloatFactory = Decimal
+except:
+    defaultFloatFactory = float
```
altendky: A bit of redundancy here reassigning `defaultFloatFactory = float`.  Also, importing * is pretty hazardous as far as name collisions and hard to trace.  Maybe:

```python3
try:
    import decimal
except ImportError:
    defaultFloatFactory = float
else:
    defaultFloatFactory = decimal.Decimal
```

But...  what is the scenario where `decimal` is appropriately not available?  It seems it was added in Python 2.4.

canmatrix/canmatrix.py
```diff
@@ -239,8 +246,8 @@ class Signal(object):
 
 
     name = attr.ib(default = "")
-    float_factory = attr.ib(default=float)
-    float_factory = float
+#    float_factory = attr.ib(default=defaultFloatFactory)
+    float_factory = defaultFloatFactory
```
altendky: Why didn't the `attr.ib()` one get kept instead?

canmatrix/canmatrix.py
```diff
@@ -536,60 +534,27 @@ class Frame(object):
...
-            ))
-
-        if self._j1939_pgn is not None:
-            self.recalcJ1939Id()
-        # we got a j1939 PGN,
 
+    name = attr.ib(default="")
+    Id = attr.ib(default = 0)
+    size = attr.ib(default = 0)
+    transmitter = attr.ib(default =[])
```
altendky: Don't use mutables as defaults, just like you shouldn't do it as regular function defaults.

https://repl.it/@altendky/why-not-mutable-parameter-defaults

```python3
transmitter = attr.ib(factory=list)
```
This will call `list` each time a new instance is created such that each one gets a unique list.

Side note, this used to be written as `attr.ib(default=attr.Factory(list))`.  Just FYI in case you see it around.

Be sure to look below at the other `[]` and `{}` defaults.

altendky:
Following the discussion in the attrs PR I think we should just from `float_factory` from the `Signal` class.  I had thought it was used for something other than just constructing but that isn't the case (at least not now).  Instead we should just pass in the value with the desired type when building a signal.  I think the basic use is loading from files where we should really use the proper developer/user-selected factory at the point where the value is parsed from the text.  If we come up with a reason that this should be handled somewhere close to `Signal` rather than at the parsing then we could use the `@classmethod` alternate constructor approach.

canmatrix/canmatrix.py
```diff
@@ -234,75 +106,48 @@ class Signal(object):
             _multiplex ('Multiplexor' or Number of Multiplex)
     """
 
-    def __init__(self, name, **kwargs):
-        self.mux_val = None
-        self.is_multiplexer = False
+    name = attr.ib(default = "")
+#    float_factory = attr.ib(default=defaultFloatFactory)
```
ebroecker: @altendky  as you mentioned in #163,
should I just remove this complete here?

remove all float_factorys here in canmatrix.py

What to do with defaults than?
Or just use Decimal everywhere. Does Decimal have any thrawbacks?

canmatrix/canmatrix.py
```diff
@@ -29,12 +29,20 @@
...
+import attr
 
 from collections import OrderedDict
 
 import logging
 import fnmatch
 
+try:
+    import decimal
+except ImportError:
```
altendky: In what scenario do we expect `decimal` to not be importable?  It's in stdlib for Python 2.4+ and it seems all version of 3.

canmatrix/canmatrix.py
ebroecker: What about making decimal to the default for all format-modules also.
One could add an option to change the behaviour. 
But are there any Problems expected while using decimal instead of float?
Wouldn't it be the best to habe decimal everywhere in canmatrix?

What do you think?

altendky: I assume there could be some unexpected behavior for people that want to use float but I haven't tried to dig into the details.  Generally though stuff like how decimal accentuates the oddities of float.

```python3
>>> decimal.Decimal(0.1)
Decimal('0.1000000000000000055511151231257827021181583404541015625')
```
But, I don't recall offhand where something like this would happen in canmatrix.  Mostly we would be creating decimals from text which works nicely.

Sorry I don't have a solid answer for this one.  :|  It seems like it would be fine to go pure decimal but I can't claim there won't be any valid complaints.

canmatrix/xls.py
```diff
@@ -538,9 +539,9 @@ def load(file, **options):
             (mini, maxi) = test.strip().split("..", 2)
             unit = ""
             try:
-                newSig.offset = float(mini)
-                newSig.min = float(mini)
-                newSig.max = float(maxi)
+                newSig.offset = float_factory(mini)
```
altendky: I didn't explore if it was actually wrong, but this looks not obviously correct at least.  The assignment of min to offset, that is.

canmatrix/dbc.py
```diff
@@ -454,7 +456,7 @@ class FollowUps(object):
         l = line.strip()
         if l.__len__() == 0:
             continue
-        try:
+#        try:
```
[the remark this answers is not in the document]
ebroecker: I just comment the try in again, it's a mistake that it's commented out...

canmatrix/xls.py
ebroecker: this is a mistake - quite sure

altendky:
I'm confused...  I can't even find a reference to `attr` when I search the changes.  Also, Github does a terrible job with reviews.  I can't even figure out what my last comment applies to.  **sigh**  but somewhere...

I feel like there should be a verification that it's actually an integer.  Old code did that (`'1.1'` would fail) but also rejected stuff like `'1e1'` that apparently needs to be supported (who writes an integer like that?  `:[`).  As is, the new code would silently truncate `'1.1'` to `1`.

Anyways, sure, drop it in dev (after solving conflicts `:]`).

PR #164, opened 2018-07-04
Move the setup_logger call to __main__ section

pierreluctg:
This will stop creating a log handler when importing from canmatrix

### DIFF
--- a/canmatrix/convert.py
+++ b/canmatrix/convert.py
@@ -25,7 +25,6 @@ from __future__ import absolute_import
 
 #from .log import setup_logger, set_log_level
 from canmatrix.log import setup_logger, set_log_level
-logger = setup_logger('root')
 import sys
 import os
 sys.path.append('..')
@@ -392,4 +391,5 @@ def main():
     convert(infile, outfileName, **cmdlineOptions.__dict__)
 
 if __name__ == '__main__':
+    logger = setup_logger('root')
     sys.exit(main())
--- a/canmatrix/convert.py
+++ b/canmatrix/convert.py
@@ -25,6 +25,8 @@ from __future__ import absolute_import
 
 #from .log import setup_logger, set_log_level
 from canmatrix.log import setup_logger, set_log_level
+import logging
+logger = logging.getLogger('root')
 import sys
 import os
 sys.path.append('..')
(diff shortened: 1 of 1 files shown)


pierreluctg:
@altendky, I fully agree with you that the logger should be `logger = logging.getLogger(__name__)`.

However, this change here still applies since the:

> Application code should not directly instantiate and use instances of Handler.
https://docs.python.org/3/howto/logging.html#handlers

So `import canmatrix` should not create a log handler.

I also wanted to make this change as small as possible to avoid breaking anything for other.

altendky:
Maybe call `setup_logger()` inside main?  I think that'd be required due to the way the entry point is specified, otherwise it won't happen unless you call the file directly (which we mostly shouldn't).

https://github.com/ebroecker/canmatrix/blob/82ece227cd2b5a7ced2e0f75610708ff1231ca04/setup.py#L87

Personally, I'd probably drop the `if __name__ ...` stuff, but that'd certainly belong in a separate commit/PR.

altendky:
@pierreluctg, I didn't mean to hold this up much.  Just that from my reading `setup_logger()` presently wouldn't be called when running from our entry point which directly calls `canmatrix.convert.main()`.  Just move `setup_logger()` inside `main()` and it'll read good to me.  Or, tell me I'm wrong and it works fine as is...  :]

canmatrix/convert.py
```diff
@@ -25,7 +25,8 @@
 
 #from .log import setup_logger, set_log_level
 from canmatrix.log import setup_logger, set_log_level
-logger = setup_logger('root')
+import logging
+logger = logging.getLogger('root')
```
altendky: Maybe just:

```python3
logger = logging.getLogger(__name__)
```
So this file gets its own logger normally regardless of being imported or run.  Then below we configure the root logger which we don't use directly.

https://docs.python.org/3/library/logging.html#logger-objects

canmatrix/convert.py
```diff
@@ -387,6 +388,7 @@ def main():
         # only print error messages, ignore verbosity flag
         verbosity = -1
 
+    setup_logger('root')
```
altendky: With how `setup_logger()` is implemented now this should probably be `setup_logger(None)`.

`setup_logger()` could be adjusted to default to `None` but this can be left for another PR, or included.  Whatever.

https://docs.python.org/3/library/logging.html#logging.getLogger

So we configure the level for the root logger below maybe this becomes:

```python3
root_logger = setup_logger(None)
set_log_level(root_logger, verbosity)
```

ebroecker:
@pierreluctg 
do you plan to resolve the last change requests, so that we can merge this PR ?

eduard

altendky:
Addressed in #237/#244 (hopefully...)

PR #176, opened 2018-08-07
Use the default value if a frame attribute is None

msalau:
Frame attributes are not verified to not to be None during initialization.
The issue was found when I tried to convert a dbf file into dbc format. The convert tool crashed in the following way:
```
msalau@msalau:~/canmatrix$ git log -1 --oneline
ce143d5 add doc to xlsx template (#165)
msalau@msalau:~/canmatrix$ python2 -m canmatrix.convert ./1.dbf ./1.dbc
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/msalau/canmatrix/canmatrix/convert.py", line 396, in <module>
    sys.exit(main())
  File "/home/msalau/canmatrix/canmatrix/convert.py", line 393, in main
    convert(infile, outfileName, **cmdlineOptions.__dict__)
  File "/home/msalau/canmatrix/canmatrix/convert.py", line 199, in convert
    canmatrix.formats.dumpp(outdbs, outfileName, **options)
  File "canmatrix/formats.py", line 113, in dumpp
    dump(db, fileObject, exportType, **options)
  File "canmatrix/formats.py", line 89, in dump
    moduleInstance.dump(canMatrixOrCluster, fileObject, **options)
  File "canmatrix/dbc.py", line 177, in dump
    if frame.transmitter.__len__() == 0:
AttributeError: 'NoneType' object has no attribute '__len__'
msalau@msalau:~/canmatrix$
```

I've tracked down the issue and discovered that the transmitter attribute is initialized with None on line 218 of dbf.py. The value then caused the crash during generation of the output file, since there is no `__len__` function associated.

The fix works for me.

Regards,
Maksim Salau.

### DIFF
--- a/canmatrix/canmatrix.py
+++ b/canmatrix/canmatrix.py
@@ -572,6 +572,8 @@ class Frame(object):
         for arg_name, destination, function, default in args:
             try:
                 value = kwargs[arg_name]
+                if value is None:
+                    value = default
             except KeyError:
                 value = default
             else:


altendky:
I'm not at my computer at the moment to dig myself...  But where are we passing `None`?  Perhaps that is the error.

msalau:
Hi @altendky 

Here is the source of None: [dbf.py#L215](https://github.com/ebroecker/canmatrix/blob/development/canmatrix/dbf.py#L215)
```
                if len(temparray) > 6:
                    transmitter = temparray[6].split()
                else:
                    transmitter = None
                newBo = db.frames.addFrame(
                    Frame(name,
                          Id=int(Id),
                          dlc=size,
                          transmitter=transmitter))
```

altendky:
Thanks for humoring me and providing the source. 
 I'm honestly not certain what is best as I have some conflicting thoughts on the issue.  But, as an alternative solution...

```python
extras = {}
if len(temparray) > 6:
    extras['transmitter'] = temparray[6]
newBo = db.frames.addFrame(
                    Frame(name,
                          Id=int(Id),
                          dlc=size,
                          **extras,
```

We'll see what @ebroecker says and I'll see if I can settle my internal dilemmas on this.  @msalau, do you have any preference?  If so, particular reasons?

I'll note that the mechanism being changed in this PR is hopefully being replaced as seen in the attrs branch.  I proposed the pattern before I learned about attrs and it's basically a bad start at what that library provides.  So, there's some consideration as to compatibility with that rework which I think tends towards what I have above but I'll have to verify.

I supposed in this particular case there's option three as well.  It should give either element 6 if it exists or an empty list if it doesn't.  This is the cleanest call but I don't really like picking 'defaults' outside a function so to speak.

```python
transmitter=temparray[6:7]
```

msalau:
Hi @altendky and @ebroecker 
I'd suggest to fix the `__init__` function, not the caller. This way we will have an explicit way to initialize the list to its default value, which is much simpler than playing with `kwargs` to my mind.
The bug exists for a long time and either was undetected, or ignored by end-users. Possibly they just used another tool in that case. By fixing the caller, we will fix this particular case, but by fixing `__init__`, we'll eliminate the possibility for such bugs to creep into the code.

Also, the current attribute naming is a little bit misleading. It is named `transmitter` but should be initialized with a list. IMHO, it is a good idea to make the function accept strings also.
This way the usage of the class will be a little bit more intuitive and robust. E.g.:
```
Frame(..., transmitter="NodeA") # transmitter = ["NodeA"]
# or 
Frame(..., transmitter=["NodeB", "NodeC"]) # transmitter = ["NodeB", "NodeC"]
# or
Frame(..., transmitter=None) # transmitter = default -> transmitter = []
```
I can make a PR, if you agree with the feature.

ebroecker:
I tend to fix this in the importer (dbf).
One point is, that I even don't know how to fix this in attr-branch...
Maybe in attr-branch one could force an error, if 'None' instead of a list is given.

altendky:
@ebroecker we could do some goofy converter maybe that changes `None` into the default but...  meh.

Here's my set of possibilities.  I'm going to run it by `#python` to see if I can get any helpful opinions or other options.

https://repl.it/@altendky/using-a-callables-default-value-1

@msalau, you'll note that `attrs` doesn't naturally work when passing in `None`.  Also, that technique doesn't generally apply in cases where `None` is a valid value but not the default.  Or where you want it invalid initially and will fill it in later.  Or...

I don't even know what `transmitter` is for, but sure, it sounds like a single item, not a list.

ebroecker:
transmitter - is the sender of the Frame. 
Shure usually it's exactly one ECU. 
But sometimes there are Frames which can be sent from multiple ECUs...

ebroecker:
You are right, fixed the typo in attr branch:
https://github.com/ebroecker/canmatrix/commit/81a40ad621d1778b6284faed10f6c6b906b23cda

ebroecker:
@msaula 
I think I fixed it already:
https://github.com/ebroecker/canmatrix/blob/development/canmatrix/dbf.py#L218

maybe you could verify?
Thanks a lot

msalau:
@ebroecker 
I worked with the dbf file that uncovered the issue. Thanks!
`canmatrix` doesn't check for version of `atts` in run-time and thus failed on first attempt. I had `atts` version 15 installed. I've updated the module and it worked fine.

What about the `master` branch?

PR #183, opened 2018-08-23
Determine the signals sign (case insensitive) and format dbc generated max to show exponent (as needed)

MartyTemple:
Hopefully, these changes are complete enough to merge in. I noticed that the Development branch has the same case sensitivity (must be lower case) for determining the signal's sign.

Thank you.


canmatrix/dbc.py
```diff
@@ -54,7 +54,7 @@ def format_float(f):
     s = str(f).upper()
     if s.endswith('.0'):
         s = s[:-2]
-
+        s = '%.7g' % float(s)
```
altendky: Given that everything is (usually?) a decimal, perhaps we could format this stuff directly as desired rather than back and forth through string and float?

MartyTemple: I will remove this from my pull request. Formatting the max value for a signal in the DBC output may not be required. I can't reproduce the tool issue that I experienced earlier. 

A little over zealous on this part of my request.

test/test.json
```diff
@@ -1,35 +1,35 @@
...
+                }
+            ]
+        },
+        {
+            "id": 18,
+            "isextended_frame": true,
+            "name": "extendedFrame",
+            "signals": []
+        }
+    ]
```
altendky: Hmm, is this just newline changes?  It would be nice if github showed a bit better...

ebroecker:
@MartyTemple 
Thanks for your PR.
Could you please have a look on the error travis complains about?
Probably you will need some sort of `baseType = None` somewhere above ...

Thanks

altendky:
Oops, now that's some timing...

Issue #146, opened 2018-04-05
Attributes


MichaelRSG:
Hi,

is it possible to add some attributes to the signals like cycle time in the Interaction Layer?

Thank you :)

altendky:
I'll note that this is available for sym format, but not as a general canmatrix feature yet I don't believe.

ebroecker:
currently the cycle time and the send type are in frame and signal attributes (frame.attributes and signal.attributes) only.

I want these both as direct properties (signal.cycleTime, signal.sendType, frame.cycleTime, frame.sendType).

ebroecker:
Physically (on CAN) a cycletime does only make sense on frames.
But if you move signals from one frame to another it is interesting to know the cycletime needs from the signal. 
In dbc it is possible to define cycletimes for signals. 
Frames can inherit the cycletimes from the signals (and use the fastest one) automatically.

I'll implement some frame properties which optionally inherit the cycletimes from the signals.
Hope this makes this issue a bit more clear.

PR #197, opened 2018-09-27
T196 fix additional frame

Funth0mas:
Extend `--additionalFrameAttributes` to support any attribute name. Add this feature to json export. Replace `eval()` by `getattr()`.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -40,8 +40,8 @@ import logging
 import fnmatch
 
 import decimal
-
 defaultFloatFactory = decimal.Decimal
+
 
 logger = logging.getLogger('root')
 try:
@@ -52,7 +52,6 @@ except:
 
 from past.builtins import basestring
 import copy
-
 
 class BoardUnit(object):
     """
@@ -72,6 +71,7 @@ class BoardUnit(object):
                 define = db.buDefines[attributeName]
                 return define.defaultValue
 
+
     def addAttribute(self, attribute, value):
         """
         adds some Attribute to current Boardunit/ECU
@@ -86,7 +86,6 @@ class BoardUnit(object):
 
     def __str__(self):
         return self.name
-
 
 def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
@@ -106,49 +105,48 @@ class Signal(object):
             _multiplex ('Multiplexor' or Number of Multiplex)
     """
 
-    name = attr.ib(default="")
-    #    float_factory = attr.ib(default=defaultFloatFactory)
+    name = attr.ib(default = "")
+#    float_factory = attr.ib(default=defaultFloatFactory)
     float_factory = defaultFloatFactory
     startBit = attr.ib(type=int, default=0)
-    size = attr.ib(type=int, default=0)
-    is_little_endian = attr.ib(type=bool, default=True)
-    is_signed = attr.ib(type=bool, default=True)
-    offset = attr.ib(converter=float_factory, default=float_factory(0.0))
-    factor = attr.ib(converter=float_factory, default=float_factory(0.0))
+    size = attr.ib(type=int, default = 0)
+    is_little_endian = attr.ib(type=bool, default = True)
+    is_signed = attr.ib(type=bool, default = True)
+    offset = attr.ib(converter = float_factory, default = float_factory(0.0))
+    factor = attr.ib(converter = float_factory, default = float_factory(0.0))
 
     #    offset = attr.ib(converter = float_factory, default = 0.0)
 
-    min = attr.ib(converter=float_factory)
-
+    min  = attr.ib(converter=float_factory)
     @min.default
     def setDefaultMin(self):
-        return self.calcMin()
+        return  self.calcMin()
 
-    max = attr.ib(converter=float_factory)
-
+    max =  attr.ib(converter = float_factory)
     @max.default
     def setDefaultMax(self):
-        return self.calcMax()
+        return  self.calcMax()
 
-    unit = attr.ib(type=str, default="")
-    receiver = attr.ib(default=attr.Factory(list))
-    comment = attr.ib(default=None)
-    multiplex = attr.ib(default=None)
+    unit = attr.ib(type=str, default ="")
+    receiver = attr.ib(default = attr.Factory(list))
+    comment = attr.ib(default = None)
+    multiplex  = attr.ib(default = None)
 
-    mux_value = attr.ib(default=None)
+    mux_value = attr.ib(default = None)
     is_float = attr.ib(type=bool, default=False)
-    enumeration = attr.ib(type=str, default=None)
-    comments = attr.ib(type=dict, default=attr.Factory(dict))
-    attributes = attr.ib(type=dict, default=attr.Factory(dict))
-    values = attr.ib(type=dict, convert=normalizeValueTable, default=attr.Factory(dict))
-    calc_min_for_none = attr.ib(type=bool, default=True)
-    calc_max_for_none = attr.ib(type=bool, default=True)
-    muxValMax = attr.ib(default=0)
-    muxValMin = attr.ib(default=0)
-    muxerForSignal = attr.ib(type=str, default=None)
+    enumeration = attr.ib(type=str, default = None)
+    comments = attr.ib(type=dict, default = attr.Factory(dict))
+    attributes = attr.ib(type=dict, default = attr.Factory(dict))
+    values = attr.ib(type=dict, convert=normalizeValueTable, default = attr.Factory(dict))
+    calc_min_for_none = attr.ib(type=bool, default = True)
+    calc_max_for_none = attr.ib(type=bool, default = True)
+    muxValMax = attr.ib(default = 0)
+    muxValMin = attr.ib(default = 0)
+    muxerForSignal= attr.ib(type=str, default = None)
 
     def __attrs_post_init__(self):
         self.multiplex = self.multiplexSetter(self.multiplex)
+
 
     @property
     def spn(self):
@@ -307,14 +305,14 @@ class Signal(object):
                     break
             else:
                 raise ValueError(
-                    "{} is invalid value choice for {}".format(value, self)
+                        "{} is invalid value choice for {}".format(value, self)
                 )
 
         if not (self.min <= value <= self.max):
             logger.info(
                 "Value {} is not valid for {}. Min={} and Max={}".format(
                     value, self, self.min, self.max)
-            )
+                )
         raw_value = (value - self.offset) / self.factor
 
         if not self.is_float:
@@ -339,7 +337,6 @@ class Signal(object):
 
     def __str__(self):
         return self.name
-
 
 class SignalGroup(object):
     """
@@ -374,13 +371,13 @@ class SignalGroup(object):
     def __iter__(self):
         return iter(self.signals)
 
-
 @attr.s(cmp=False)
 class Frame(object):
     """
     Contains one CAN Frame.
 
-    The Frame has  following mandatory attributes\n
+    The Frame has  following mandatory attributes
+
     * id,
     * name,
     * transmitters (list of boardunits/ECU-names),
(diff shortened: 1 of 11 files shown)


altendky:
Yeah, `eval()` is...  ugly.  But, I'm not a huge fan of making dict keys available as attributes either.  The cli effect could be achieved without that as well.  I can't say I have a really strong argument, it mostly just seems less 'magical'.  (and yes, I have worked with `__getattr__` etc, so I don't mean magical like I am not familiar with it)

```python
getattr(signal, item, signal.attributes.get(item, ""))
```

Or just add a getter.

```python
NOTHING = object()
def get(self, item, default=NOTHING):
    item = getattr(self, item, self.attributes.get(item, default))

    if item is NOTHING:
        raise SomeException()

    return item
```
```python
signal.get(item, "")
```

Funth0mas:
Totally agree, your solution is better. I'll apply it.

There is still question if it make sense to have some "additional attributes" as instance variables and some in the attributes dictionary. It would make sense to have them all on a single place. But I didn't want to touch the existing code, because...  I'm just a passerby.

altendky:
If I remember a conversation with @ebroecker correctly, it would be nice to revisit these 'attributes' and try to figure out which are common and how to nestle the rest together happily.  Basically, there is a dbc bias and with the extra knowledge of other formats we could probably do a better job picking regular attributes to have.

Thanks for seeing the mess and doing something about at least part of it.  :]  We appreciate the help.

ebroecker:
@Funth0mas 
having all attributes in one place is not that easy...
My intention was to have the "common" attributes direct in the frame (yes some are missing). 
The attributes[]-Dictionary is more or less a "dbc-only" place to save all dbc-attributes. The problem of dbc is, that you can define any attribute in dbc. 
Some attributes are not yet in the common list (#146)

src/canmatrix/canmatrix.py
```diff
@@ -425,62 +436,75 @@ def pgn(self, value):
...
+
+    def attribute(self, name, db=None, default=None):
+        """Get any Frame attribute by its name.
+
+        :param str name: attribute name, can be mandatory (ex: id) or optional (customer) attribute.
+        :param CanMatrix db: Optional database parameter to get global default attribute value.
+        :param default: Default value if attribute doesn't exist.
+        :return any: Return the attribute value if found, else `default` or None
+        """
+        if hasattr(self, name):
```
altendky: Sorry to evolve my request here...  I forgot this was an `attrs` class and I think this will be a touch better.  It will avoid considering methods and such as attributes.

https://www.attrs.org/en/stable/api.html#attr.fields_dict

```python
if name in attr.fields_dict(type(self)):
```

I'll note that this is technically a breaking change given the it will block access via this method to `.attribute` items that shadow the names of regular attributes.  But, I don't think that bothers me.

src/canmatrix/canmatrix.py
```diff
@@ -425,62 +436,75 @@ def pgn(self, value):
...
+    #    def cycleTime(self, value):
+    #        self._cycleTime = value
+    #        self.attributes["GenMsgCycleTime"] = value
+
+    #    @sendType.setter
+    #    def sendType(self, value):
+    #       self._sendType = value
+    #       self.attributes["GenMsgCycleTime"] = value
+
+    def attribute(self, name, db=None, default=None):
```
altendky: This is definitely a breaking change though it does seem better.  I think it's better for `canmatrix` to be tending towards improving so I'm comfortable with this but figured I'd highlight it.

src/canmatrix/canmatrix.py
```diff
@@ -425,62 +436,75 @@ def pgn(self, value):
...
+            return self.attributes[name]
         elif db is not None:
-            if attributeName in db.frameDefines:
-                define = db.frameDefines[attributeName]
+            if name in db.frameDefines:
+                define = db.frameDefines[name]
                 return define.defaultValue
         else:
-            return None
+            return default
```
altendky: +1

src/canmatrix/cmjson.py
```diff
@@ -101,21 +101,21 @@ def dump(db, f, **options):
                               "signals": signals}
             frame_attributes = {}
             for frame_info in additionalFrameColums:  # Look for additional Frame Attributes
-                if hasattr(frame, frame_info):
-                    frame_attributes[frame_info] = getattr(frame, frame_info)
+                if frame.attribute(frame_info):  # does the attribute exist?
```
altendky: This doesn't check it's existence.  If it exists but is `None` or anything else falsey it will be treated as not present.  Here's the first thing that came to my mind, though it is a touch smelly to me.  Maybe you can think of something nicer.

```python
MISSING = object()
value = frame.attribute(name=frame_info, default=MISSING)
if value is not MISSING:
    frame_attributes[frame_info] = value
```

src/canmatrix/cmjson.py
```diff
@@ -101,21 +101,21 @@ def dump(db, f, **options):
...
             if frame_attributes:  # only add Attributes if there are any
                 symbolic_frame["attributes"] = frame_attributes
             exportArray.append(symbolic_frame)
-    else:  # exportall
+    else:  # exportAll
         for frame in db.frames:
             frameattribs = {}
             for attribute in db.frameDefines:
-                frameattribs[attribute] = frame.attribute(db,attribute)
+                frameattribs[attribute] = frame.attribute(attribute, db)
```
altendky: Let's specify by name to avoid this hazard in the future.  Looks like this applies to most of the rest of the diff below.

Funth0mas:
I ran the autoformating on the whole file by accident. If it's problem, I'll revert it line per line. Sorry for the inconvenience. 
Mainly I removed the cryptic getattr and I propose to use the existing attribute() getter, but make the db parameter really optional. I added some docstrings (only in Frame) when trying to understand the code.

altendky:
No big deal, I just tend to mention more rather than less.  Thanks for the help improving the code.

src/canmatrix/canmatrix.py
```diff
@@ -370,45 +374,52 @@ def __str__(self):
...
-    transmitters (list of boardunits/ECU-names),
-    size (= DLC),
-    signals (list of signal-objects),
-    attributes (list of attributes),
-    receiver (list of boardunits/ECU-names),
-    extended (Extended Frame = 1),
-    comment
+    Contains one CAN Frame.
+
+    The Frame has  following mandatory attributes\n
```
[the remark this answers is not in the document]
Funth0mas: There must be either empty line or \n befor the list (ctrl+Q in PyCharm). Changed to empty line.

src/canmatrix/canmatrix.py
```diff
@@ -105,49 +106,50 @@ class Signal(object):
             _multiplex ('Multiplexor' or Number of Multiplex)
     """
 
-    name = attr.ib(default = "")
-#    float_factory = attr.ib(default=defaultFloatFactory)
+    name = attr.ib(default="")
+    #    float_factory = attr.ib(default=defaultFloatFactory)
```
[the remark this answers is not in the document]
Funth0mas: (comic is that it was done automatically by PyCharm) ... clear, reverted, the same for all another unwanted changes. Comments, indentation and another codestyle issues can be fixed later in own PR.

src/canmatrix/canmatrix.py
Funth0mas: If this change is accepted, I'd adopt the similar functions of another classes accordingly.

src/canmatrix/cmjson.py
Funth0mas: good point! As there is imho no reason to export None value, change to `is not None`

src/canmatrix/cmjson.py
Funth0mas: Uff, I realized the PyCharm has refactored also calls of another classes. Python has certain weaknesses :-(
Checked, reverted unwanted changes. For Frame use named attribute.

src/canmatrix/canmatrix.py
```diff
@@ -780,7 +816,8 @@ def __init__(self, definition):
         def saveConvertStrToInt(inStr):
             out = int(defaultFloatFactory(inStr))
             if out != defaultFloatFactory(inStr):
-                logger.warning("Warning, integer was expected but got float: got: {0} using {1}\n".format(inStr, str(out)))
+                logger.warning(
+                    "Warning, integer was expected but got float: got: {0} using {1}\n".format(inStr, str(out)))
```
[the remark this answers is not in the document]
Funth0mas: unwanted change, revert.

Funth0mas:
Ready for review again. Thanks!

src/canmatrix/canmatrix.py
```diff
@@ -409,6 +416,7 @@ class Frame(object):
 
     @property
     def is_multiplexed(self):
+        """Frame is multiplexed if at least one of its signals is multiplexed."""
```
ebroecker: Frame is multiplexed it at least one of its signals is a multiplexer

src/canmatrix/canmatrix.py
ebroecker: canmatrix is at 0. - version. I agree that improving is always better than staying compatible.  
I'd accept it!

src/canmatrix/canmatrix.py
```diff
@@ -552,45 +584,54 @@ def globSignals(self, globStr):
...
         """
-        add attribute to attribute-list of frame; If Attribute already exits, modify value
+        Add the attribute with value to customer Frame attribute-list. If Attribute already exits, modify its value.
+        :param str attribute: Attribute name
+        :param any value: attribute value
         """
         try:
             self.attributes[attribute] = str(value)
-        except:
+        except:  # could it really fail?
```
ebroecker: I'm not sure why I added this try-catch. But I'm quite sure that there was a problem without it.
Maybe there is a problem with "str" if your are in python2 and have some issues with charsets...

src/canmatrix/cmcsv.py
```diff
@@ -183,10 +183,7 @@ def dump(db, thefile, delimiter=',', **options):
 
         additionalFrameInfo = []
         for frameInfo in additionalFrameCollums:
-            try:
-                temp = eval("frame." + frameInfo)
-            except:
-                temp = ""
+            temp = frame.attribute(frameInfo, default="")
```
ebroecker: +++ this eliminates my dirty "eval"-hack.
Thanks

src/canmatrix/canmatrix.py
altendky: None of this should be dealt with in this PR but:

1) it should catch the particular exception not just anything
2) i think users should be expected to pass the proper type
3) i think there are still various str-or-int-or-... around so maybe it would get caught up in cleaning that up.

src/canmatrix/cmcsv.py
```diff
@@ -270,3 +261,7 @@ def dump(db, thefile, delimiter=',', **options):
 #        finalTableString = "\n".join(
 #            [row.toCSV(delimiter) for row in csvtable])
 #        print(finalTableString)
+    if sys.version_info > (3, 0):
+        # When TextIOWrapper is garbage collected, it closes the raw stream
+        # unless the raw stream is detached first
+        temp.detach()
```
altendky: Funny thing, I just ran into this over in a tqdm PR.  Sensible addition given the `if` above but the thing overall sort of smells to me.  `*hopes he wasn't the one that added it*`

Should probably really be in a finally clause though just in case there are exceptions.  Or a context manager, whatever you feel like.

Funth0mas: I simply copied it from another module because there were an error/warning after running the Unit Test informing about unreleased resource. To be honest I don't know why the `TextIOWrapper` must be present. Either let it be for further cleanup (I can create a ticket) or I need more time to investigate. The first solution is currently more convenient for me :-)

src/canmatrix/canmatrix.py
Funth0mas: I'll refactor the `attribute()` of another classes to have the same signature in another PR (when I don't fall asleep in the train next time)

Funth0mas:
The failed appveyor isn't my fault  :-(

src/canmatrix/cmcsv.py
altendky: Yeah, all I mean for this is to make a try/finally to be sure the detach actually gets run even in case of exceptions.

Funth0mas:
Hi, I merged the current origin/development and now it seems I did more changes than I really did, when looking to "changed files". I'm used to do this with bitbucket but it has different result. Did I something wrong?

altendky:
@Funth0mas are you referring to 3bdafa9?  I generally do a `git merge --no-ff development` and get a merge commit.  I'm not really sure quite why it was diffing the way it was...  but I think I just fixed it.  Fwiw, I did a `git rebase -i HEAD~3` and deleted the squash commit then did a regular `git merge development --no-ff` and a force push to your branch (you let us have write access to just that branch when you submitted the PR).  Yes, I did also push a backup branch to my repo first.  :]

Anyways, if this looks right to you I'm ready to merge.

Funth0mas:
@altendky funny thing is that I did almost the same except of using more complicated `git reset` and force pushed two minutes before you did :-) At least I understand the git a little bit more now. 
I 'inspected' the diff once again and yes, it's the state I wanted to have. Thank you for your attention.

altendky:
Thanks for all the code and your responsiveness.

Cheers,
-kyle

Funth0mas:
My pleasure. BTW you're from Germany?
Regards

ebroecker:
No, only I am from Germany, altendkey Not...

PR #202, opened 2018-10-09
Update attribute() signature and add/update some docstrings.

Funth0mas:
As promissed, I unified the signature of attribute() functions.
Added docstrings to canmatrix.py.
Fixed few typos.
Add some remarks (comments ending with question mark)

Sorry to mix again these change types but it was single-pass work. I still believe it's reviewable, I strongly talked to my fingers not to do _all_ they wanted to :-) . I hope it helps to increase the code lucidity.

Thx for review.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -63,24 +63,36 @@ class BoardUnit(object):
         self.attributes = dict()
         self.comment = None
 
-    def attribute(self, db, attributeName):
+    def attribute(self, attributeName, db=None, default=None):
+        """Get Board unit attribute by its name.
+
+        :param str attributeName: attribute name.
+        :param CanMatrix db: Optional database parameter to get global default attribute value.
+        :param default: Default value if attribute doesn't exist.
+        :return: Return the attribute value if found, else `default` or None
+        """
         if attributeName in self.attributes:
             return self.attributes[attributeName]
-        else:
+        elif db is not None:
             if attributeName in db.buDefines:
                 define = db.buDefines[attributeName]
                 return define.defaultValue
-
+        return default
 
     def addAttribute(self, attribute, value):
         """
-        adds some Attribute to current Boardunit/ECU
+        Add the Attribute to current Boardunit/ECU. If the attribute already exists, update the value.
+
+        :param str attribute: Attribute name
+        :param any value: Attribute value
         """
         self.attributes[attribute] = value
 
     def addComment(self, comment):
         """
-        Set comment of Signal
+        Set Board unit comment.
+
+        :param str comment: BU comment/description.
         """
         self.comment = comment
 
@@ -94,15 +106,18 @@ def normalizeValueTable(table):
 @attr.s(cmp=False)
 class Signal(object):
     """
-    contains on Signal of canmatrix-object
-    with following attributes:
-            name, startBit,size (in Bits)
-            is_little_endian (1: Intel, 0: Motorola)
-            is_signed ()
-            factor, offset, min, max
-            receiver  (Boarunit/ECU-Name)
-            attributes, _values, _unit, _comment
-            _multiplex ('Multiplexor' or Number of Multiplex)
+    Represents a Signal in CAN Matrix.
+
+    Signal has following attributes:
+
+    * name
+    * startBit, size (in Bits)
+    * is_little_endian (1: Intel, 0: Motorola)
+    * is_signed (bool)
+    * factor, offset, min, max
+    * receiver  (Boarunit/ECU-Name)
+    * attributes, _values, unit, comment
+    * _multiplex ('Multiplexor' or Number of Multiplex)
     """
 
     name = attr.ib(default = "")
@@ -112,8 +127,8 @@ class Signal(object):
     size = attr.ib(type=int, default = 0)
     is_little_endian = attr.ib(type=bool, default = True)
     is_signed = attr.ib(type=bool, default = True)
-    offset = attr.ib(converter = float_factory, default = float_factory(0.0))
     factor = attr.ib(converter = float_factory, default = float_factory(1.0))
+    offset = attr.ib(converter = float_factory, default = float_factory(0.0))
 
     #    offset = attr.ib(converter = float_factory, default = 0.0)
 
@@ -150,6 +165,7 @@ class Signal(object):
 
     @property
     def spn(self):
+        """Get signal J1939 SPN or None if not defined."""
         return self.attributes.get("SPN", None)
 
     def multiplexSetter(self, value):
@@ -158,62 +174,86 @@ class Signal(object):
         if value is not None and value != 'Multiplexor':
             ret_multiplex = int(value)
             self.mux_val = int(value)
-        else:
+        else:  # is it valid for None too?
             self.is_multiplexer = True
             ret_multiplex = value
         return ret_multiplex
 
-    def attribute(self, db, attributeName):
+    def attribute(self, attributeName, db=None, default=None):
+        """Get any Signal attribute by its name.
+
+        :param str attributeName: attribute name, can be mandatory (ex: startBit, size) or optional (customer) attribute.
+        :param CanMatrix db: Optional database parameter to get global default attribute value.
+        :param default: Default value if attribute doesn't exist.
+        :return: Return the attribute value if found, else `default` or None
+        """
+        if attributeName in attr.fields_dict(type(self)):
+            return getattr(self, attributeName)
         if attributeName in self.attributes:
             return self.attributes[attributeName]
-        else:
+        if db is not None:
             if attributeName in db.signalDefines:
                 define = db.signalDefines[attributeName]
                 return define.defaultValue
+        return default
 
     def addComment(self, comment):
         """
-        Set comment of Signal
+        Set signal description.
+
+        :param str comment: description
         """
         self.comment = comment
 
     def addReceiver(self, receiver):
         """
-        add receiver Boardunit/ECU-Name to signal
+        Add signal receiver.
+
+        :param BoardUnit receiver: Boardunit/ECU instance.
         """
         if receiver not in self.receiver:
             self.receiver.append(receiver)
 
     def delReceiver(self, receiver):
         """
-        remove receiver Boardunit/ECU-Name from signal
+        Remove receiver (Boardunit/ECU) from signal
+
+        :param BoardUnit receiver: Boardunit/ECU instance.
         """
         if receiver in self.receiver:
             self.receiver.remove(receiver)
 
     def addAttribute(self, attribute, value):
         """
-        Add Attribute to Signal
+        Add user defined attribute to the Signal. Update the value if the attribute already exists.
+
+        :param str attribute: attribute name
+        :param value: attribute value
         """
         self.attributes[attribute] = value
 
     def delAttribute(self, attribute):
         """
-        Remove Attribute to Signal
-        """
+        Remove user defined attribute from Signal.
 
+        :param str attribute: attribute name
+        """
         if attribute in self.attributes:
             del self.attributes[attribute]
 
     def addValues(self, value, valueName):
         """
-        Add Value/Description to Signal
+        Add named Value Description to the Signal.
+
+        :param int value: signal value (0xFF)
+        :param str valueName: Human readable value description ("Init")
         """
         self.values[int(value)] = valueName
 
     def setStartbit(self, startBit, bitNumbering=None, startLittle=None):
         """
-        set startBit.
+        Set startBit.
+
         bitNumbering is 1 for LSB0/LSBFirst, 0 for MSB0/MSBFirst.
         If bit numbering is consistent with byte order (little=LSB0, big=MSB0)
         (KCD, SYM), start bit unmodified.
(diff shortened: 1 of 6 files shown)


src/canmatrix/xls_common.py
```diff
@@ -84,17 +84,17 @@ def getSignal(db, sig, motorolaBitFormat):
     # startvalue of signal available
     if "GenSigStartValue" in db.signalDefines:
         if db.signalDefines["GenSigStartValue"].definition == "STRING":
-            frontArray.append(sig.attribute(db,"GenSigStartValue"))
+            frontArray.append(sig.attribute("GenSigStartValue"), db=db)
```
altendky: Need to shift a paren outside the `db=db.  Presently it has `db` as a parameter to `.append()`.

src/canmatrix/canmatrix.py
```diff
@@ -94,15 +106,18 @@ def normalizeValueTable(table):
...
+    Signal has following attributes:
+
+    * name
+    * startBit, size (in Bits)
+    * is_little_endian (1: Intel, 0: Motorola)
+    * is_signed (bool)
+    * factor, offset, min, max
+    * receiver  (Boarunit/ECU-Name)
+    * attributes, _values, unit, comment
+    * _multiplex ('Multiplexor' or Number of Multiplex)
```
altendky: Is this a standard doc form?  I can't say I've done much with docs but if we are filling this stuff in it might be worth finding doc generation that can work with attrs.

src/canmatrix/canmatrix.py
```diff
@@ -338,27 +396,40 @@ def raw2phys(self, value, decodeToStr=False):
...
+
 class SignalGroup(object):
     """
-    contains Signals, which belong to signal-group
+    Represents signal-group, containing multiple Signals.
     """
 
     def __init__(self, name, Id):
-        self.signals = []
+        self.signals = []  # use attr.ib?
```
altendky: Separate PR, but sure, go for it.

src/canmatrix/canmatrix.py
```diff
@@ -338,27 +396,40 @@ def raw2phys(self, value, decodeToStr=False):
...
+        """Remove Signal from SignalGroup.
+
+        :param Signal signal: signal to remove
+        """
         if signal in self.signals:
-            self.signals[signal].remove()
+            self.signals.remove(signal)
 
-    def byName(self, name):
+    def byName(self, name):  # add 'Signal' to function name?
```
altendky: It's a signal group, what would you get other than signals?  I mean, maybe, but that's my gut response.  :]

src/canmatrix/canmatrix.py
```diff
@@ -517,10 +596,12 @@ def addSignalGroup(self, Name, Id, signalNames):
             if signalId is not None:
                 newGroup.addSignal(signalId)
 
-    def signalGroupbyName(self, name):
+    def signalGroupbyName(self, name):  # capital B: ByName?
```
altendky: I would say yes.

Though I would `snake_case` anyways.  That'd be a separate issue I presume.

src/canmatrix/canmatrix.py
```diff
@@ -853,8 +951,8 @@ def saveConvertStrToInt(inStr):
         elif definition[0:3] == 'HEX':
             self.type = 'HEX'
             min, max = definition[4:].split(' ', 2)
-            self.min = saveConvertStrToInt(min)
-            self.max = saveConvertStrToInt(max)
+            self.min = safeConvertStrToInt(min)  # the same for INT and HEX?
```
altendky: I agree, that doesn't sound right.  But, I don't know offhand what triggers this code.

src/canmatrix/canmatrix.py
```diff
@@ -1119,11 +1287,18 @@ def recalcDLC(self, strategy):
...
-        if type(old).__name__ == 'instance':
+        """Rename ECU in the Matrix. Update references in all Frames.
+
+        :param str or BoardUnit old: old name or ECU instance
+        :param str newName: new name
+        """
+        if type(old).__name__ == 'instance':  # accept BoardUnit instance
             pass
-        else:
+        else:  # or string
```
altendky: Yeah, we need to stop doing this type guessing stuff.  I'd say either `renameEcuByBoardUnit()` and `renameEcuByName()` or we just check if old is in `self.boardUnits` and not then treat it as a string.  But, this is a separate issue.

Not to mention that when we do check type it should be `isinstance(this, That)`. :]  Looks like there are a few places to apply this.

src/canmatrix/canmatrix.py
```diff
@@ -1227,43 +1421,77 @@ def delSignal(self, signal):
...
+        :param BoardUnit ecu: Receiver ECU name
+        """
+        frames = self.globFrames(framePattern)
         for frame in frames:
             signal = frame.signalByName(signalName)
             if signal is not None:
                 signal.addReceiver(ecu)
                 for signal in frame.globSignals(signalName):
-                    signal.addReceiver(ecu)
+                    signal.addReceiver(ecu)  # already set, redundant line
```
altendky: Huh.  This seems very explicit but yeah, I'm not sure the cases where `frame.globSignals(signalName)` would get additional signals.  @ebroecker?  See similar below as well.

src/canmatrix/canmatrix.py
```diff
@@ -1278,9 +1506,10 @@ def merge(self, mergeArray):
                         "Name Conflict, could not copy/merge EnvVar " + envVar)
 
     def setFdType(self):
```
altendky: This sounds more like `guessIfFdAndSet()` or somesuch.  @ebroecker, what's the history on this?  We can't tell if a frame is FD or not?  Or can you not have an FD frame with <= 8 bytes?  Why only setting to `True` and no case for setting `False`?

src/canmatrix/canmatrix.py
ebroecker: Oh ... YES. I just forgot SignalGroup while switching to attr.ib

src/canmatrix/canmatrix.py
ebroecker: I fully agree

src/canmatrix/canmatrix.py
```diff
@@ -1136,24 +1311,26 @@ def renameEcu(self, old, newName):
             frame.updateReceiver()
 
     def addEcu(self, ecu):
+        """Add new ECU to the Matrix. Do nothing if ecu with the same name already exists.
```
ebroecker: I mixed up Ecu and BU (=BoardUnit) --
One name should be enough.
Which one we should choose?

src/canmatrix/canmatrix.py
```diff
@@ -1174,8 +1352,13 @@ def updateEcuList(self):
...
 
     def renameFrame(self, old, newName):
+        """Rename Frame.
+
+        :param Frame or str old: Old Frame instance or name or part of the name with '*' at the beginning or the end.
+        :param str newName: new Frame name, suffix or prefix
+        """
         if type(old).__name__ == 'instance':
-            oldName = old.name
+            old = old.name
```
ebroecker: + you found a bug!

src/canmatrix/canmatrix.py
ebroecker: I could imagine, that the indentation is wrong here.
The Idea was to check first if there is a signal with the exact name.
Afterwards check if there is a signal searching with globSignals.

But this makes no sense.
There should only be the "globSignals" way to find the right signals, the "signalByName" could be removed?!

src/canmatrix/canmatrix.py
```diff
@@ -1227,43 +1421,77 @@ def delSignal(self, signal):
...
+        """
+        frames = self.globFrames(framePattern)
         for frame in frames:
             signal = frame.signalByName(signalName)
             if signal is not None:
-                signal.addReceiver(ecu)
+                signal.delReceiver(ecu)
                 for signal in frame.globSignals(signalName):
-                    signal.delReceiver(ecu)
+                    signal.delReceiver(ecu)  # redundant
```
ebroecker: I think same copy paste issue from addSignalReceiver

src/canmatrix/canmatrix.py
ebroecker: you are right - this is only guessing.
Meanwhile with J1939 this may even do mistakes.
In deed I think FD has long DLCs only - but I may be wrong. If I am wrong, setting False would be wrong also :-)

src/canmatrix/xls_common.py
ebroecker: You are right, this must be fixed

src/canmatrix/canmatrix.py
Funth0mas: Yes. Try ctrl+Q in pycharm, it gets rendered.

altendky: I use pycharm, but it wouldn't be my primary target when picking a doc format.

src/canmatrix/canmatrix.py
Funth0mas: It might be hard job, the refactoring capabilities of pycharm aren't unlimited as I saw :-( But yes, the fingers want to write snake case.

src/canmatrix/canmatrix.py
Funth0mas: Mainly I did not remarked the 2nd `addReceiver` runs on another object with the same name. I'll try to thing about it once againe.

src/canmatrix/canmatrix.py
Funth0mas: :-) No idea but I hope the sphinx can be converted to another format "automatically"

src/canmatrix/canmatrix.py
Funth0mas: Personally I've never heard 'Board Unit' but always 'ECU'. But it doesn't matter. The abbreviation 'ECU' seems clearer than 'BU' for me. You decide :-)

src/canmatrix/canmatrix.py
Funth0mas: I thing the data speed depends merely on the flag in the CAN Frame header on the bus. That means "DLC>8 implies CAN FD" but DLC <= 8 implies nothing.

The function could than rather be `checkFdRequired()`. And set back to False, if there is nothing with long DLC. Agree?

src/canmatrix/canmatrix.py
Funth0mas: But than, in code, you see `return tmp.byName(name)` and you're lost... The __getitem__() might be aslo useful. Postpone to another PR.. maybe

src/canmatrix/canmatrix.py
Funth0mas: Leaving the code untouched for further refactoring.

src/canmatrix/canmatrix.py
ebroecker: OK, ECU should be the name for it in future.

src/canmatrix/canmatrix.py
ebroecker: Agree

ebroecker:
is there anything open, or could we merge?

src/canmatrix/canmatrix.py
Funth0mas: I'll propose it separately so that this "big" PR contains minimum changes.

src/canmatrix/canmatrix.py
```diff
@@ -158,70 +174,93 @@ def multiplexSetter(self, value):
         if value is not None and value != 'Multiplexor':
             ret_multiplex = int(value)
             self.mux_val = int(value)
-        else:
+        else:  # is it valid for None too?
```
Funth0mas: this is still unclear but it might be resolved in the future

Funth0mas:
I checked the diff once again and I don't see any issue which should be part of this PR. Rather for further work.

Which concerns the docstring format, they should be convertible by pyment.

src/canmatrix/canmatrix.py
altendky: I wouldn't use method names to help callers keep track of what sort of object they have.  It seems like the name referring to the object itself is what should be sufficiently descriptive for the context.

src/canmatrix/canmatrix.py
```diff
@@ -1227,43 +1417,71 @@ def delSignal(self, signal):
...
-    def addSignalReceiver(self, frameName, signalName, ecu):
-        frames = self.globFrames(frameName)
+    def addSignalReceiver(self, framePattern, signalName, ecu):
+        """Add Receiver to all Frames by name pattern.
+
+        :param str framePattern: Frame name pattern
+        :param str signalName: signal name
+        :param str ecu: Receiver ECU name
+        """
+        frames = self.globFrames(framePattern)
```
altendky: If it's going to be exclusively a glob I'd use that in the name rather than the generic 'pattern'.  Ditto for the docstring.  Glob, like regex, isn't a single thing but it's a _relatively_ standard pattern so I'd think it's useful to be explicit.

May as well catch `signalName` as well if we're going to exclusively glob it.  Also, find all similar below (or move them to their own PR).

src/canmatrix/canmatrix.py
altendky: As is this could probably be at least 3 I think.  :]  Every new change that's added holds up all the rest from being accepted.

src/canmatrix/canmatrix.py
altendky: What has a class name of (lower-case) `instance` anyways?

Funth0mas:
The code with 'instance' is not mine. For the rest - please excuse my anarchism, I'll revert all issues you mentioned back to the previous version. I'll be more careful next time. Thank you for your patience.

PR #203, opened 2018-10-10
Try to make Unit Tests alive

Funth0mas:
I'm trying to fix unit tests and get some reasonable coverage report. I want to see the result on the build server. (locally on windows it already seems better)

No review is necessary now. I just want to trigger the build server.


Funth0mas:
Thanks :-) Does the codecov run new tests?  Or does the travis ini "tox -e codecov" mean "exclude"? (Ok I can google it on my own...)

altendky:
https://github.com/ebroecker/canmatrix/blob/30ae7e22a549a26da34031bcef17fe969b5aa152/tox.ini#L17

That's the line that is running the new tests, including measuring coverage.

test/test_codec.py
```diff
@@ -24,175 +27,175 @@ def test_bitstruct_format(self):
...
         self.assertEqual(s6.raw2phys(5), 10)
         self.assertEqual(s6.raw2phys(255), s6.max)
         self.assertEqual(s6.raw2phys(0), s6.min)
 
-        s7 = Signal('signal', signalSize=16, is_signed=False, factor=0.001)
+        s7 = Signal('signal', size=16, is_signed=False, factor=0.001)
         self.assertEqual(s7.raw2phys(65535), s7.max)
         self.assertEqual(s7.raw2phys(0), s7.min)
-        self.assertEqual(s7.raw2phys(50123), 50.123)
+        self.assertAlmostEqual(s7.raw2phys(50123), D(50.123))
```
altendky: Shouldn't using decimal get us exact?  But, definitely don't `decimal.Decimal(1.23)` but rather `decimal.Decimal('1.23')` (note the latter is a string).  I see this is closed, but still a relevant FYI on using decimal.

PR #204, opened 2018-10-10
Fix jsonAll, add UT for it.

Funth0mas:
Fix jsonAll, wrap by try/finally, add UT

### DIFF
--- a/test/test_codec.py
+++ b/test/test_codec.py
@@ -205,3 +205,19 @@ class TestCanmatrixCodec(unittest.TestCase):
             with open(out_file_name, "r") as file:
                 data = file.read()
             self.assertIn("UserFrameAttr", data)
+
+    def test_export_jsonall(self):
+        """Check the jsonAll doesn't raise and export some additional field."""
+        db_path = os.path.join(
+            os.path.dirname(__file__), "..", "test", "test.dbc")
+        dbs = formats.loadp(db_path)
+        tmp_dir = tempfile.mkdtemp()
+        out_file_name = tmp_dir + "/output.json"
+        formats.dumpp(dbs, out_file_name, jsonAll=True)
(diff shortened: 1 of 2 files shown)


test/test_codec.py
```diff
@@ -205,3 +205,19 @@ def test_import_export_additional_frame_info(self):
             with open(out_file_name, "r") as file:
                 data = file.read()
             self.assertIn("UserFrameAttr", data)
+
+    def test_export_jsonall(self):
+        """Check the jsonAll doesn't raise and export some additional field."""
+        db_path = os.path.join(
+            os.path.dirname(__file__), "..", "test", "test.dbc")
```
altendky: Is it plausible to make a minimal example 'file' here (`io.BytesIO`) that covers just this case?  Or maybe it doesn't even need to be a file and could just be built up in a few lines from raw canmatrix objects.

test/test_codec.py
```diff
@@ -205,3 +205,19 @@ def test_import_export_additional_frame_info(self):
...
+        dbs = formats.loadp(db_path)
+        tmp_dir = tempfile.mkdtemp()
+        out_file_name = tmp_dir + "/output.json"
+        formats.dumpp(dbs, out_file_name, jsonAll=True)
```
altendky: I should know this for sure...  but I _think_ we could just dump it to a string?  :[ I really ought to know this offhand.  If nothing else, I'm pretty sure we could dump it into an `io.BytesIO`.

test/test_codec.py
Funth0mas: Built up from raw canmatrix objects.

test/test_codec.py
Funth0mas: BytesIO applied

PR #205, opened 2018-10-10
Properly use 'decimal' in UTs

Funth0mas:
When playing with (yes, old) tests I found out some problems when mixing `decimal.Decimal` with `floats` ~~in single arithmetic operation. Also the converting back to int must be `round`, because `int(4.99999)` equals to 4, what is (I believe, and the old UTs also) not the wanted result.~~

The invalid (failing) tests are commented out for now.

I'll copy/move the reasonable old test to new test folder once I get a bit familiar with the pytest.

Regards

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -349,7 +349,7 @@ class Signal(object):
     def phys2raw(self, value=None):
         """Return the raw value (= as is on CAN).
 
-        :param decimal.Decimal or str value: (scaled) value or value choice to encode
+        :param value: (scaled) value compatible with `decimal` or value choice to encode
         :return: raw unscaled value as it appears on the bus
         :rtype: int or decimal.Decimal
         """
@@ -380,7 +380,7 @@ class Signal(object):
     def raw2phys(self, value, decodeToStr=False):
         """Decode the given raw value (= as is on CAN).
 
-        :param int or decimal.Decimal value: raw value. Don't use the type 'float' to remain accuracy.
+        :param value: raw value compatible with `decimal`.
         :param bool decodeToStr: If True, try to get value representation as *string* ('Init' etc.)
         :return: physical value (scaled)
         """
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -351,7 +351,7 @@ class Signal(object):
 
         :param decimal.Decimal or str value: (scaled) value or value choice to encode
         :return: raw unscaled value as it appears on the bus
-        :rtype: int or float
+        :rtype: int or decimal.Decimal
         """
         if value is None:
             return int(self.attributes.get('GenSigStartValue', 0))
@@ -374,13 +374,13 @@ class Signal(object):
         raw_value = (value - self.offset) / self.factor
 
         if not self.is_float:
-            raw_value = round(raw_value)
-        return float(raw_value)
+            raw_value = int(raw_value)
+        return raw_value
 
     def raw2phys(self, value, decodeToStr=False):
         """Decode the given raw value (= as is on CAN).
 
-        :param value: raw value. Preferably int or str, not float (to not lose precision)
+        :param int or decimal.Decimal value: raw value. Don't use the type 'float' to remain accuracy.
         :param bool decodeToStr: If True, try to get value representation as *string* ('Init' etc.)
         :return: physical value (scaled)
         """
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -349,7 +349,7 @@ class Signal(object):
     def phys2raw(self, value=None):
         """Return the raw value (= as is on CAN).
 
-        :param value: (scaled) value or value choice to encode
+        :param decimal.Decimal or str value: (scaled) value or value choice to encode
         :return: raw unscaled value as it appears on the bus
         :rtype: int or float
         """
@@ -366,8 +366,6 @@ class Signal(object):
                         "{} is invalid value choice for {}".format(value, self)
                 )
 
-        if not isinstance(value, defaultFloatFactory):  # use the same data type as offset and factor
-            value = defaultFloatFactory(value)
         if not (self.min <= value <= self.max):
             logger.info(
                 "Value {} is not valid for {}. Min={} and Max={}".format(
@@ -380,12 +378,13 @@ class Signal(object):
         return float(raw_value)
 
     def raw2phys(self, value, decodeToStr=False):
-        """Decode the given raw value (= as is on CAN)
-        :param value: raw value
+        """Decode the given raw value (= as is on CAN).
+
+        :param value: raw value. Preferably int or str, not float (to not lose precision)
         :param bool decodeToStr: If True, try to get value representation as *string* ('Init' etc.)
         :return: physical value (scaled)
         """
-        value = defaultFloatFactory(value)
+
         value = value * self.factor + self.offset
         if decodeToStr:
             for value_key, value_string in self.values.items():
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -351,6 +351,7 @@ class Signal(object):
 
         :param value: (scaled) value or value choice to encode
         :return: raw unscaled value as it appears on the bus
+        :rtype: int or float
         """
         if value is None:
             return int(self.attributes.get('GenSigStartValue', 0))
@@ -365,6 +366,8 @@ class Signal(object):
                         "{} is invalid value choice for {}".format(value, self)
                 )
 
+        if not isinstance(value, defaultFloatFactory):  # use the same data type as offset and factor
+            value = defaultFloatFactory(value)
         if not (self.min <= value <= self.max):
             logger.info(
                 "Value {} is not valid for {}. Min={} and Max={}".format(
@@ -373,8 +376,8 @@ class Signal(object):
         raw_value = (value - self.offset) / self.factor
 
         if not self.is_float:
-            raw_value = int(raw_value)
-        return raw_value
+            raw_value = round(raw_value)
+        return float(raw_value)
 
     def raw2phys(self, value, decodeToStr=False):
         """Decode the given raw value (= as is on CAN)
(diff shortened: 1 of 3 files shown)


src/canmatrix/canmatrix.py
```diff
@@ -316,16 +319,16 @@ def phys2raw(self, value=None):
         raw_value = (value - self.offset) / self.factor
 
         if not self.is_float:
-            raw_value = int(raw_value)
-        return raw_value
+            raw_value = round(raw_value)
+        return float(raw_value)
```
altendky: If we want to convert to a not-int then wouldn't a float factory be the way?  Specifically `self.float_factory` here it seems.  Though I'm thinking this is not needed as `self.offset` and `self.factor` should already be the proper types and as such `raw_value` would generally be the proper type already.

src/canmatrix/canmatrix.py
```diff
@@ -308,6 +309,8 @@ def phys2raw(self, value=None):
                         "{} is invalid value choice for {}".format(value, self)
                 )
 
+        if not isinstance(value, defaultFloatFactory):  # use the same data type as offset and factor
+            value = defaultFloatFactory(value)
```
altendky: I'm trying to work us away from this.  The caller should pass the proper thing rather than us guessing what the proper thing is.  In particular here, this check doesn't work for a `defaultFloatFactory` that isn't a class.

test/test_codec.py
```diff
@@ -5,10 +5,13 @@
 import unittest
 import tempfile
 import os
-from canmatrix import formats
+import decimal
 
+from canmatrix import formats
 from canmatrix.canmatrix import Signal
 
+D = decimal.Decimal
```
altendky: We can see what @ebroecker says but I'm not a fan of this.  It makes it basically impossible to search for or see uses of the thing at a glance.

test/test_codec.py
```diff
@@ -24,175 +27,175 @@ def test_bitstruct_format(self):
...
         self.assertEqual(s8.phys2raw(), 5)
 
-        s9 = Signal('signal', signalSize=16, is_signed=False, factor=0.001)
+        s9 = Signal('signal', size=16, is_signed=False, factor=0.001)
         self.assertEqual(s9.phys2raw(), 0)
-        self.assertEqual(s9.phys2raw(s9.max), 65534)
+        self.assertEqual(s9.phys2raw(s9.max), 65535)
         self.assertEqual(s9.phys2raw(s9.min), 0)
-        self.assertEqual(s9.phys2raw(50.123), 50123)
+        self.assertEqual(s9.phys2raw(D(50123)*D(0.001)), 50123)
```
altendky: As noted before, pass a string to `decimal.Decimal` instead of a float to avoid loss of accuracy.

src/canmatrix/canmatrix.py
Funth0mas: Okay but this works , the defaultFloatFactory  is `decimal.Decimal` so it's a class. But it can become a function one day - I understand.
The problem is, that if we pass an `int` to the function, it produces _a result_, but (slightly) wrong result, instead of raising an exception.

test/test_codec.py
Funth0mas: I am that lazy :-( I'll get rid of it.

src/canmatrix/canmatrix.py
altendky: Sorry, I see where you were going and yes, it makes sense.  Forget my comment.  Though, we are stuck without separate single and double float types.  I think I have gotten to see a couple float signals in the wild actually... Right before they got converted to fixed point.

Now that I think about it round doesn't actually give you an int, does it?

src/canmatrix/canmatrix.py
altendky: I'm admittedly pretty tired right now...  Could you provide an example where the original code got an integer input wrong?

src/canmatrix/canmatrix.py
Funth0mas:
```python
>>> type(round(4.999))
<class 'int'>
```
But I don't have python 2 to check it there.

src/canmatrix/canmatrix.py
Funth0mas: When being tired by my faulty PR, don't hesitate to say it :-D

I wrote nonsense. Regarding phys2raw the `int` works and `float` raises `TypeError('unsupported operand type(s) for -:')`. Maybe try/catch + raise with better explanation. Or simply let it be.

src/canmatrix/canmatrix.py
ebroecker: for me it's ok, if py3 works. 
Py2 is not primary development platform anymore. 
As I am not sure phys2raw will stay, because something is wrong with decoding/encoding anyway. So maybe API for decoding/encoding changes in future.

Funth0mas:
I hope so :-)

src/canmatrix/canmatrix.py
altendky: I'd say we fully support py2 until we drop it entirely.  Unless we are going to pull a maintenance-only py2 branch, but even then the py2 branch should be fully py2 correct.  I don't think either is happening yet so let's not muddy the waters.  And yes, in py2 `round()` returns a float.  Easy enough to deal with in this case.

https://docs.python.org/2/library/functions.html#round
> The result is a floating point number.

src/canmatrix/canmatrix.py
altendky: Interesting, so it does.  Yeah, I think I'd leave it for now.  It tells you the line and the two incompatible types.  Either the caller just forgot or they probably ought to learn about the issues with floats.  *shrug*

On the reraise topic, in py3 be sure to `raise DescriptiveError() from original_exception` so the original isn't lost.  Sadly there's no py2/3 compatible solution that isn't totally custom so we don't get to do that.  Leaves me more hesitant to make 'helpful' replacement exceptions.

src/canmatrix/canmatrix.py
altendky: I'm still hesitant about the float thing.  It loses information (the exact value).  Of course, what's proper totally depends on the point of this function.  Is it to give you a type whose representation is exactly what would be on the bus?  Or to just handle the scaling?  Should it even be doing rounding vs. truncation?  So I guess I'm not sure what it should be but I'd like it consistent.  If we're going to float (which still misses single vs double precision) then it seems we ought to be returning an int for integer signals.  We can leave the bigger picture debates for later.

Funth0mas:
Please don't merge yet.

Funth0mas:
Finally there are no code changes in `phys2raw` and `raw2phys`. I was misleaded by "old" tests which obvoiusly were written before the `Decimal` was introduced. I wrote all fraction values as str (as well when instantiation the Signal) and all tests passes as expected. Thank you for correcting me.

Summary
- docstrings correction, no code change
- update old tests
- "move" old tests for p2r and r2p to the new test directory (is it ok to have them in test_canmatrix.py?)

altendky:
Oh, and update the PR title.  :]

test/test_codec.py
```diff
@@ -24,175 +25,175 @@ def test_bitstruct_format(self):
...
-            data = {
-                'Signal': 2,
-                'someTestSignal': 101,
-            }
-            data_bytes = tuple(bytearray(bus.encode(test_frame1, data)))
-            assert data_bytes == (0, 0x28, 0x04, 0, 0, 0, 0, 0)
+        self.assertEqual(s10.phys2raw(Decimal('0.005')), 100)
+        self.assertEqual(s10.phys2raw(Decimal('0.003')), 60)
+
+    # def test_encode_canmatrix(self):
```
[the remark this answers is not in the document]
Funth0mas: I mentionned in original PR description, that some failing tests are commented out, because they fail and probably need rework (I'd check them later, if you don't forbid me to continue...). Probably bad idea again :-(

test/test_codec.py
```diff
@@ -24,175 +25,175 @@ def test_bitstruct_format(self):
...
-        s4 = Signal('signal', signalSize=8, is_little_endian=False)
+        s4 = Signal('signal', size=8, is_little_endian=False)
         self.assertEqual(s4.bitstruct_format(), '>s8')
 
     def test_encode_signal(self):
-        s1 = Signal('signal', signalSize=8)
+        s1 = Signal('signal', size=8)
         self.assertEqual(s1.phys2raw(), 0)
-        self.assertEqual(s1.phys2raw(1), 1)
+        self.assertEqual(s1.phys2raw(Decimal(1)), 1)
```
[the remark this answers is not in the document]
Funth0mas: It's residuum from when I tried to understand what's wrong, before you explained me, that floats must be written as str. I'll fix it.

src/canmatrix/canmatrix.py
```diff
@@ -377,8 +378,9 @@ def phys2raw(self, value=None):
         return raw_value
 
     def raw2phys(self, value, decodeToStr=False):
-        """Decode the given raw value (= as is on CAN)
-        :param value: raw value
+        """Decode the given raw value (= as is on CAN).
+
+        :param int or decimal.Decimal value: raw value. Don't use the type 'float' to remain accuracy.
```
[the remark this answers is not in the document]
Funth0mas: I wanted to say "first learn about decimal and than use this"... for those like me. Possibly I'd omit it here and write some note to `Signal` docstring.

test/test_codec.py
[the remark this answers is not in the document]
altendky: Was it just annoying sifting through test results with failures you didn't care about?  'Silencing' them permanently isn't great.  I won't hold up a PR based on existing test failures that still fail (well, not intentionally).  I probably will on commenting out tests so the failures aren't reported.  Unless maybe they somehow cause problems beyond a failed message.

test/test_codec.py
```diff
@@ -24,72 +25,72 @@ def test_bitstruct_format(self):
...
 
-        s3 = Signal('signal', signalSize=8, is_signed=False)
+        s3 = Signal('signal', size=8, is_signed=False)
         self.assertEqual(s3.bitstruct_format(), '<u8')
 
-        s4 = Signal('signal', signalSize=8, is_little_endian=False)
+        s4 = Signal('signal', size=8, is_little_endian=False)
         self.assertEqual(s4.bitstruct_format(), '>s8')
 
     def test_encode_signal(self):
```
altendky: I just realized that you copied as opposed to moving the test.  Do move instead, so delete this copy I think.  Ditto for decode.  That'll make it easier to keep track of what is left in the old test world.

PR #209, opened 2018-10-15
attr.ibify SignalGroup

Funth0mas:
Use attr in SignalGroup. Add tests.

I propose to access signals by [] operator as well.

Thank you for review.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -403,7 +403,7 @@ class SignalGroup(object):
     """
     name = attr.ib(type=str)
     id = attr.ib(type=int)
-    signals = attr.ib(type=list, default=attr.Factory(list), repr=False)
+    signals = attr.ib(factory=list, repr=False)
 
     def addSignal(self, signal):
         """Add a Signal to SignalGroup.
@@ -442,7 +442,7 @@ class SignalGroup(object):
         signal = self.byName(name)
         if signal:
             return signal
-        raise IndexError("Signal {} doesn't exist".format(name))
+        raise KeyError("Signal '{}' doesn't exist".format(name))
 
 
 @attr.s(cmp=False)
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -396,15 +396,14 @@ class Signal(object):
         return self.name
 
 
+@attr.s(cmp=False)
 class SignalGroup(object):
     """
     Represents signal-group, containing multiple Signals.
     """
-
-    def __init__(self, name, Id):
-        self.signals = []
-        self.name = name
-        self.id = Id
+    name = attr.ib(type=str)
+    id = attr.ib(type=int)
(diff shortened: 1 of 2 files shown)


ebroecker:
@Funth0mas 
Thanks again for your work!

There is no way to chat directly in github...
So sometimes it is more easy to communicate outside of github.
With altendky I sometimes communicate via hangouts or email.
Maybe - if you are interested - send me a email [my address is in my github profile].

src/canmatrix/canmatrix.py
```diff
@@ -435,13 +434,16 @@ def byName(self, name):
...
-
     def __iter__(self):
         """Iterate over all contained signals."""
         return iter(self.signals)
 
+    def __getitem__(self, name):
+        signal = self.byName(name)
+        if signal:
+            return signal
+        raise IndexError("Signal {} doesn't exist".format(name))
```
altendky: `KeyError`?  Though it's not really either so maybe a custom exception.

Maybe `repr(name)` or otherwise get it in quotes?

src/canmatrix/tests/test_canmatrix.py
```diff
@@ -11,3 +12,22 @@ def test_signal_defaults_to_decimal():
 
     assert isinstance(signal.offset, decimal.Decimal)
     assert isinstance(signal.factor, decimal.Decimal)
+
+
+# SignalGroup
+def test_sinalgroup_add_remove():
```
altendky: Missing the 'g' in `sinal`.

I'd have a separate test for each feature.  I see add signal adds, add signal doesn't duplicate, group is iterable, `.byName()` finds something, get item access, `.byName()` finds nothing, deleting existent works, deleting not existent doesn't raise, invalid get item access raises.  Separating them out gives more resolution when looking at the failure reports rather than having to read the actual exception and only knowing the first failed test.

I do see that I cheated and checked two things in my sample test.  :|  I'm not a stickler for exactly one assert per test, but I'd tend away from one test for a whole class at least.

src/canmatrix/canmatrix.py
altendky: Yeah, I believe I know experienced people that prefer both ways.  I like the separation so callers can make more specific except clauses.  I might change it later, but I'll let `KeyError` go for now.  :]

Funth0mas:
Unfortunately my "help" means that you spend more time by review than writing it on your own :-/

altendky:
Sometimes yes, sometimes no.  With time the ratio generally gets better.  FWIW, I go through this myself with my PR's elsewhere.  Other eye's look with a different perspective which can be quite helpful.  I do appreciate your help and am glad you push through despite all my complaints.  :]

Funth0mas:
Resolved conflict. Travis failed due to network issue. I've probably no rights to rerun it.

altendky:
A close and reopen of the PR will trigger a rebuild when you don't have rights at the CI.  Or a `git commit --allow-empty 'kick ci'`.  But I got it in this case.

PR #211, opened 2018-10-17
[WIP] Fix frame encoding

Shevonar:
Decimal is the default float-factory since e8fdd35a which means that float values cannot be
encoded because - operator for float and Decimal is incompatible in Signal.phys2raw().

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -175,7 +175,7 @@ class Signal(object):
             ret_multiplex = int(value)
             self.mux_val = int(value)
         else:  # is it valid for None too?
-            self.is_multiplexer = True
+            self.is_multiplexer = False
             ret_multiplex = value
         return ret_multiplex
 
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -870,7 +870,7 @@ class Frame(object):
                     signals = sorted(self.signals, key=lambda s: s.getStartbit())
                     signals_values = OrderedDict()
                     for signal, value in zip(signals, bitstruct.unpack(fmt, data)):
-                        signals_values[signal.name] = signal.raw2phys(value, decodeToStr)
+                        signals_values[signal.name] = signal.raw2phys(defaultFloatFactory(value), decodeToStr)
                 muxVal = int(signals_values.values()[0])
             # find all signals with the identified multiplexer-value
             muxedSignals = []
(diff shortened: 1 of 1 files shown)


altendky:
In my mind the point of decimal not working with floats is so that you don't accidentally use floats.  I think we want to either have the caller explicitly convert or they could set the float factory when creating... Though I think the latter may be broken in some cases.

altendky:
Let's keep separate changes in separate PR's.  Then we won't hold up good with bad/needs-work and it's easier to keep track of reviews.

src/canmatrix/canmatrix.py
```diff
@@ -870,8 +870,8 @@ def decode(self, data, decodeToStr=False):
                     signals = sorted(self.signals, key=lambda s: s.getStartbit())
                     signals_values = OrderedDict()
                     for signal, value in zip(signals, bitstruct.unpack(fmt, data)):
-                        signals_values[signal.name] = signal.raw2phys(value, decodeToStr)
-                muxVal = int(signals_values.values()[0])
+                        signals_values[signal.name] = signal.raw2phys(defaultFloatFactory(value), decodeToStr)
```
altendky: What wasn't working that you added this?  Perhaps add a test to `src/canmatrix/tests/test_canmatrix.py` that was broken and gets fixed by the change.

Shevonar:
Yes I am sorry, I have no time to do it correctly right now. Just need stuff to work. Feel free to reject any changes. Maybe I provide good PRs later this week.

src/canmatrix/canmatrix.py
```diff
@@ -175,7 +175,7 @@ def multiplexSetter(self, value):
             ret_multiplex = int(value)
             self.mux_val = int(value)
         else:  # is it valid for None too?
-            self.is_multiplexer = True
+            self.is_multiplexer = False
```
altendky: I'm not clear why this should change.  Can you explain?

src/canmatrix/canmatrix.py
Shevonar: I added a test with more explanation in the follow up PR #215.

src/canmatrix/canmatrix.py
Shevonar: I did some more investigation and provide a better fix in PR #213.

Shevonar:
This PR is superseded by #213, #214 and #215. Sorry for all the noise, I hope the new PRs are of better quality.

PR #212, opened 2018-10-17
attr.ibify BoardUnit.

Funth0mas:
attr.ibify BoardUnit. Add UTs.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -53,15 +53,16 @@ except:
 from past.builtins import basestring
 import copy
 
+
+@attr.s
 class BoardUnit(object):
     """
     Contains one Boardunit/ECU
     """
 
-    def __init__(self, name):
-        self.name = name.strip()
-        self.attributes = dict()
-        self.comment = None
+    name = attr.ib(type=str)
+    comment = attr.ib(type=str, default=None)
+    attributes = attr.ib(factory=dict, repr=False)
 
     def attribute(self, attributeName, db=None, default=None):
         """Get Board unit attribute by its name.
@@ -96,8 +97,6 @@ class BoardUnit(object):
         """
         self.comment = comment
 
-    def __str__(self):
-        return self.name
 
 def normalizeValueTable(table):
     return {int(k): v for k, v in table.items()}
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -61,7 +61,7 @@ class BoardUnit(object):
     """
 
     name = attr.ib(type=str)
-    comment = attr.ib(type=str, default=None)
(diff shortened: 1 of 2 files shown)


src/canmatrix/canmatrix.py
```diff
@@ -53,15 +53,16 @@
...
     """
     Contains one Boardunit/ECU
     """
 
-    def __init__(self, name):
-        self.name = name.strip()
-        self.attributes = dict()
-        self.comment = None
+    name = attr.ib(type=str)
+    comment = attr.ib(type=str, default=None)
```
altendky: Shouldn't type be `typing.Optional[str]`?  Then we get to add the `typing` backport as a dependency since it is only stdlib in Python>=3.5.

Or maybe we should drop the None option and there's always a comment, it just may be an empty string?  I dunno, but that'd be a separate change anyways

Funth0mas: For this concrete case: the Signal has comment with default="". Make sense to have the same in BoardUnit too. (Moreover, I posed myself a question: does it make sense to have None as default, and allow the user to set an empty string? Is there any difference between `""` and `None` from user point of view? I mean: no.)
Generaly: as far as I understand, the attr `type` is only for static analysis and they encourage to replace it by new type hinting. We can't do that due to py2 compatibility. But there is still py2 annotation style (PEP484, PEP526). 
Would you agree with this?
```suggestion
    comment = attr.ib(default="")  # type: str
```

the functions are currently annotated in comments (okay, in docstrings). Annotation attr.ibs in comments would be more ... conform. When dropping py2 support once, it could be refactored to new annotation style by a regex.

I thing we could annotate for example in Frame as follows:
```python
receiver = attr.ib(factory=list)  # type:List[str]
```
with the `typing` dependency, obviously.
I thing the type annotations are useful for users and help to understand the code faster.

altendky: How's this.  Rather than piecemeal tweaking `None` vs. `""` here then there then somewhere else  we make a ticket.  Whoever can share opinions there and we decide what it should be everywhere and then have a PR to apply that to the entire library consistently.  This is also a distinctly breaking change so it'd probably be good to be fairly explicit about it rather than continuously evolving and breaking in different places.  (or maybe we are already more consistent than I am guessing...)

Given where we are with py2 still I think that the pep-approved comment style is fine.  But I admittedly haven't read up on it yet.

Note that even though annotations already existed, the typing module is new in 3.5 so even without py2 support we will still need the typing backport to continue 3.4 support for anything but pure types.

Funth0mas:
Okay, than I'd simply remove the 'str' annotation (the PR was meant simply to add attr.s) and create another PR with attr cleanup (use 'factory=dict' etc plus improve the type hints) And maybe two tickets for discussion.

PR #213, opened 2018-10-18
Fix signal is_multiplexer being True by default

Shevonar:
The default should be False, otherwise the first signal is encoded twice and decoding fails. I would answer the question in the code `is it valid for None too?` with No.

The error was introduced probably by always applying the `multiplexSetter` in `__attrs_post_init__` in commit a8893523.


Shevonar:
Apparently I managed to srew up this branch again. I can force push a cleaned up version if you like or undo the unwanted changes in a new commit. Which one do you prefer?

PR #214, opened 2018-10-18
Convert OrderedDict.values() to list for python3 compatibility

Shevonar:
Dict values are not index accessible in python 3. Converting to list fixes this issue.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -873,7 +873,7 @@ class Frame(object):
                     signals_values = OrderedDict()
                     for signal, value in zip(signals, bitstruct.unpack(fmt, data)):
                         signals_values[signal.name] = signal.raw2phys(value, decodeToStr)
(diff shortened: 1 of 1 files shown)


altendky:
Hopefully a temporary glitch getting pypy on AppVeyor.  Kicking another build.

Shevonar:
The simple test from #215 should execute this code. I don't know how to explicitly test for python3 compatibility. Maybe I can extend the test in #215 to cover more cases than just the one I fixed there.

altendky:
The test suite gets run against 2.7, 3.4-3.7 (plus some pypy) so any test you write will be executed against all of them.  You only need to explicitly test any particular version for a feature we only provide in a single version. 
 I don't think we intentionally have that anywhere.

Anyways, we can do #215 first and a catchup merge for this PR will get the test and should report coverage on the diff.  Thanks.

altendky:
Consumed by #215

Issue #216, opened 2018-10-21
Raise exceptions instead of just logging errors


altendky:
https://github.com/ebroecker/canmatrix/blob/057b4848dc8e3f5246111cc5d400eaa4ad2db75f/src/canmatrix/canmatrix.py#L47-L51

https://github.com/ebroecker/canmatrix/blob/057b4848dc8e3f5246111cc5d400eaa4ad2db75f/src/canmatrix/canmatrix.py#L860-L862

Should be an exception, no special message or handling, just let the import raise.

PR #217, opened 2018-10-21
Just depend on bitstruct

altendky:
#216


altendky:
It's a light dependency, why not.

Issue #207, opened 2018-10-11
Signed values are specified incorrectly in DBF writer


msalau:
Hello,

I've just discovered that during conversion from DBC to DBF format signed signals are specified in the way that `BusMaster` cannot understand correctly.

Here is the original dbc snippet:
```
BO_ 1 Frame1: 8 Vector__XXX
 SG_ Signal0 : 0|8@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Signal1 : 8|8@1- (1,0) [-2|1] "" Vector__XXX
```
Converted to dbf by `BusMaster`:
```
[START_MSG] Frame1,1,8,2,1,S,
[START_SIGNALS] Signal0,8,1,0,U,1,0,1,0,1,,,
[START_SIGNALS] Signal1,8,2,0,I,1,-2,1,0,1,,,
[END_MSG]
```
Converted to dbf by `canmatrix`:
```
[START_MSG] Frame1,1,8,2,1,S,Vector__XXX
[START_SIGNALS] Signal0,8,1,0,U,1.0,0.0,1,0.0,1.0,,,Vector__XXX
[START_SIGNALS] Signal1,8,2,0,S,1.0,-2.0,1,0.0,1.0,,,Vector__XXX
[END_MSG]
```

So the issue is in the way signess is defined. `BusMaster` uses `U` for unsigned and `I` for signed, but `canmatrix` uses `U` for unsigned `S` for signed.

The offending line is:
https://github.com/ebroecker/canmatrix/blob/3041a4e4868db4b2ba4e883fc5dd47393f52b0fe/src/canmatrix/dbf.py#L356

The issue exists only in `dump()`, `load()` considers any letter other than `U`, `F` and `D` as signed.
BTW, `BusMaster` has one more letter: `B` (Boolean?) for single bit signals.
E.g.:
```
 SG_ Signal2 : 16|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Signal3 : 16|1@1- (1,0) [0|1] "" Vector__XXX
```
turns into:
```
[START_SIGNALS] Signal2,1,3,0,B,1,0,1,0,1,,,
[START_SIGNALS] Signal3,1,3,0,B,1,0,1,0,1,,,
```
So `BusMaster` ignores sign of bit sized signals.

ebroecker:
@msalau 
Thanks for your bug report and your proposed fix.
I'll merge

PR #219, opened 2018-10-23
Fix import of signed signals from DBF files

msalau:
BusMaster uses 'I' to denote signed signals, but not 'S'.
Thus conversion from DBC to DBF format is broken, but in opposite direction works just fine.
The PR fixes issue #207


ebroecker:
THANKS

PR #221, opened 2018-10-25
Add CanId tests. 

Funth0mas:
Add CanId tests and beautify str representation, it prints
```DA:0x01 PGN:0xABCD SA:0x02```
instead of
```DA:0X1 PGN:0XABCD SA:0X2```

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -1635,5 +1635,5 @@ class CanId(object):
         return self.destination, self.pgn, self.source
 
     def __str__(self):
-        return "DA:{da:#02X} PGN:{pgn:#04X} SA:{sa:#02X}".format(
+        return "DA:0x{da:02X} PGN:0x{pgn:04X} SA:0x{sa:02X}".format(
             da=self.destination, pgn=self.pgn, sa=self.source)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/canmatrix/canmatrix.py [hunk 1]
```diff
@@ -1635,5 +1635,5 @@ def tuples(self):
         return self.destination, self.pgn, self.source
 
     def __str__(self):
-        return "DA:{da:#02X} PGN:{pgn:#04X} SA:{sa:#02X}".format(
+        return "DA:0x{da:02X} PGN:0x{pgn:04X} SA:0x{sa:02X}".format(
```
altendky: Side note:
Didn't know about this.  Though the class is entirely J1939 specific.

Funth0mas: I hope this `__str__` is intended for debugging purposes only. At least I expect the `DA:{da:#02X}` was ment to render 0x01 but the given length includes the 0X (by `#`) and leads to 0X1. I guess it should have been `DA:{da:#04X}` what would be 0X01 - still the ugly upper case X :-(

Funth0mas: ...and maybe again it doesn't make sense to test the `__str__`, I don't know.

altendky: I support the change and the test.  It's just the class name that isn't descriptive.  But, that's unrelated to this PR.

My questioning of testing the attribute setter or better wasn't actually about testing it, it was about having the setter to begin with.  Trivial functions should have tests in my opinion, even if the tests are also trivial.  Trivial regressions can happen just as easily as non-trivial ones.

PR #224, opened 2018-10-26
Fix frame decoding

ebroecker:
this is NOT ready for merge, but in a PR the could be better done.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -50,7 +50,6 @@ except:
     from itertools import izip_longest as zip_longest
 from itertools import chain
 import struct
-import bitstruct
 
 from past.builtins import basestring
 import copy
@@ -334,21 +333,16 @@ class Signal(object):
 
         return self.offset + (rawMax * self.factor)
 
-    def bitstruct_format(self):
-        """Get the bit struct format for this signal.
-
-        :return: bitstruct representation of the Signal
-        :rtype: str
-        """
-        endian = '<' if self.is_little_endian else '>'
-        if self.is_float:
-            bit_type = 'f'
-        else:
-            bit_type = 's' if self.is_signed else 'u'
-
-        return endian + bit_type + str(self.size)
-
     def unpack_bitstring(self, length, is_float, is_signed, bits):
+        """
+        returns a value calculated from bits
+        :param length: length of signal in bits
+        :param is_float: value is float
+        :param bits: value as bits (array/iterable)
+        :param is_signed: value is signed
+        :return:
+        """
+
         if is_float:
             types = {
                 32: '>f',
@@ -376,6 +370,14 @@ class Signal(object):
         return value
 
     def pack_bitstring(self, length, is_float, value, signed):
+        """
+        returns a value in bits
+        :param length: length of signal in bits
+        :param is_float: value is float
+        :param value: value to encode
+        :param signed: value is signed
+        :return:
+        """
         if is_float:
             types = {
                 32: '>f',
@@ -509,16 +511,30 @@ class SignalGroup(object):
 
 
 class decodedSignal(object):
+    """
+    Contains a decoded signal (frame decoding)
+
+    * rawValue : rawValue (value on the bus)
+    * physValue: physical Value (the scaled value)
+    * namedValue: value of Valuetable
+    * signal: pointer signal (object) which was decoded
+    """
     def __init__(self, rawValue, signal):
         self.rawValue = rawValue
         self.signal = signal
 
     @property
     def physValue(self):
+        """
+        :return: physical Value (the scaled value)
+        """
         return self.signal.raw2phys(self.rawValue)
 
     @property
     def namedValue(self):
+        """
+        :return: value of Valuetable
+        """
         return self.signal.raw2phys(self.rawValue, decodeToStr=True)
 
 
@@ -861,6 +877,13 @@ class Frame(object):
 
 
     def signals_to_bytes(self, data):
+        """Return a byte string containing the values from data packed
+        according to the frame format.
+
+        :param data: data dictionary of signal : rawValue
+        :return: A byte string of the packed values.
+        """
+
         little_bits = [None] * (self.size * 8)
         big_bits = list(little_bits)
         for signal in self.signals:
@@ -903,7 +926,6 @@ class Frame(object):
 
         :param dict data: data dictionary
         :return: A byte string of the packed values.
-        :rtype: bitstruct
         """
 
 
@@ -934,6 +956,12 @@ class Frame(object):
         return self.signals_to_bytes(data)
 
     def bytes_to_bitstrings(self, data):
+        """Return two arrays big and little containing bits of given data (bytearray)
+
+        :param data: bytearray of bits (little endian). Iterable or bytes.
+            i.e. (0xA1, 0xA2, 0xA3, 0xA4, 0xA5, 0xA6, 0xA7, 0xA8)
+        :return: bit arrays in big and little byteorder
+        """
         b = tuple('{:08b}'.format(b) for b in data)
         little = ''.join(reversed(b))
         big = ''.join(b)
@@ -941,6 +969,13 @@ class Frame(object):
         return little, big
 
     def bitstring_to_signal_list(self, signals, big, little):
+        """Return OrderedDictionary with Signal Name: object decodedSignal (flat / without support for multiplexed frames)
+
+        :param signals: Iterable of signals (class signal) to decode from frame.
+        :param big: bytearray of bits (big endian).
+        :param little: bytearray of bits (little endian).
+        :return: array with raw values (same order like signals)
+        """
         unpacked = []
         for signal in signals:
             if signal.is_little_endian:
@@ -959,6 +994,13 @@ class Frame(object):
         return unpacked
 
     def unpack(self, data, report_error=True):
+        """Return OrderedDictionary with Signal Name: object decodedSignal (flat / without support for multiplexed frames)
+
+        :param data: Iterable or bytes.
+            i.e. (0xA1, 0xA2, 0xA3, 0xA4, 0xA5, 0xA6, 0xA7, 0xA8)
+        :return: OrderedDictionary
+        """
+
         rx_length = len(data)
         if rx_length != self.size and report_error:
             print(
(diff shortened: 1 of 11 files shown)


ebroecker:
@altendky 
your unpack code worked out _out the box_ ...

Now I tried to use your _pack_ code also, but stumbled over some magic code:
https://github.com/altendky/st/blob/c75f436f503adcb4708131a65b059c8062484ae5/epyqlib/canneo.py#L136-L147
I am not sure what happens here, maybe you could give me a hint?
Thanks in advance

altendky:
Let's start with a vague overview... I'm just rereading the code since I really don't remember this stuff.

Keep in mind that I wanted to allow for overlapped signals and mixed big and little endian as well.  Unpacking overlapped signals... Whatever.  A single bit from the bus might end up in multiple signals, no conflict.  Packing though, you may have to prioritize one over another, or report a conflict (this code prioritizes I think which is probably a bad choice).

I think the first two lines take the bits calculated from little endian signals and 'turn them around' to align with the big endian bits.

`bitstring =` iterates over the little and big bits together.  The `next()` picks the first non-`None` from the choices of the little endian bit, the big endian bit, and the default `'0'`.

The `return` line groups the bits into groups of 8, joins each group to an eight character string, creates an integer from the base-2 string, and creates a bytes object from that iterable of integers.

Maybe?

src/canmatrix/dbc.py
```diff
@@ -521,7 +521,8 @@ class FollowUps(object):
                     regexp_raw = re.compile(pattern.encode(dbcImportEncoding))
                     temp = regexp.match(decoded)
                     temp_raw = regexp_raw.match(l)
-                    receiver = list(map(str.strip, temp.group(12).split(',')))
+                    receiverTemp = [str(b) for b in temp.group(12).split(',')]
```
altendky:
```suggestion
                    receiver = [b.strip() for b in temp.group(12).split(',')]
```
(and remote `receiver = ` below)

src/canmatrix/tests/test_frame_decoding.py
```diff
@@ -0,0 +1,85 @@
+import pytest
+import canmatrix.formats
+import os.path
+import sys
+
+def test_decode_from_dbc():
+    here = os.path.dirname(os.path.realpath(__file__))
+    cm = canmatrix.formats.loadp(os.path.join(here ,"test_frame_decoding.dbc"), flatImport = True)
+    # 001#8d00100100820100
+    if sys.version_info > (3, 0):
```
altendky:
```suggestion
    if sys.version_info > (3,):
```
As is it treats 3.0 as py2.  But, tests seem to pass fine if we just always use `bytearray()`.  Some reason not to drop the version check?  (consider repeats below if change is made)

src/canmatrix/dbc.py
```diff
@@ -490,7 +490,7 @@ class FollowUps(object):
                 regexp_raw = re.compile(pattern.encode(dbcImportEncoding))
                 temp_raw = regexp_raw.match(l)
                 if temp:
-                    receiver = list(map(str.strip, temp.group(11).split(',')))
+                    receiver = list(map(str.strip, [str(b) for b in temp.group(11).split(',')]))
```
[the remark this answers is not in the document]
ebroecker: didn't fell good to me also, just "try and error" with pypy.
I'm gonna try your suggestions...

src/canmatrix/tests/test_frame_decoding.py
ebroecker: good point.
I'll change to bytearray everything, than it's py2/py3 compatible.

src/canmatrix/dbc.py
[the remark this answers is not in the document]
altendky: This is a general issue so...  #233.

altendky:
I'm not sure at the moment but we should be able to monitor for unrun tests (like would happen with matching names) via coverage of the tests.

ebroecker:
@altendky 
this PR is getting close to done - in my eyes.
One open thing is to kick out bitstruct of the requirements, but the code should work now as expected...

Could you do another review?

Thanks

src/canmatrix/canmatrix.py
```diff
@@ -400,7 +400,7 @@ def pack_bitstring(self, length, is_float, value, signed):
                                signed=signed)
             b = '{:0{}b}'.format(int.from_bytes(b, byteorder='big'),
                                  length)
-            bitstring = b[:length]
```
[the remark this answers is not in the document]
altendky: Huh...  I wonder how I haven't had issues.  It seems to show up for signed values (I forget big endian or little or...).  Anyways, thanks for pointing that out.  As a double check I made up #235 to recreate it 'in my world' of `.sym` files.

src/canmatrix/canmatrix.py
```diff
@@ -779,108 +869,172 @@ def updateReceiver(self):
...
-            pass
+            Exception("encoding complex multiplexed frames not yet implemented")
         elif self.is_multiplexed:
             # search for mulitplexer-signal
             muxSignal = None
             for signal in self.signals:
                 if signal.is_multiplexer:
                     muxSignal = signal
                     muxVal = data.get(signal.name)
+                    break
```
altendky: Maybe drop the `muxSignal = None` above and have:

```python3
for signal in self.signals:
    if signal.is_multiplexer:
        muxSignal = signal
        muxVal = data.get(signal.name)
        break
else:
    raise MissingMuxSignalError(...)
```

src/canmatrix/canmatrix.py
```diff
@@ -779,108 +869,172 @@ def updateReceiver(self):
...
+            unpacked = self.bitstring_to_signal_list(self.signals, big, little)
+
+            returnDict= dict()
+
+            for s, v in zip(self.signals, unpacked):
+                returnDict[s.name] = decodedSignal(v, s)
+
+            return returnDict
+
+    def decode(self, data):
```
altendky: So, this is just `unpack()` with multiplex support?  The names don't really clarify anything to me.

src/canmatrix/canmatrix.py
```diff
@@ -779,108 +869,172 @@ def updateReceiver(self):
...
+        :param data: Iterable or bytes.
+            i.e. (0xA1, 0xA2, 0xA3, 0xA4, 0xA5, 0xA6, 0xA7, 0xA8)
+        :return: OrderedDictionary
+        """
+        decoded = self.unpack(data)
+
         if self.is_complex_multiplexed:
             # TODO
-            pass
+            raise Exception("decoding complex multiplexed frames not implemented")
```
altendky: much better :] custom exception though

src/canmatrix/tests/test_frame_decoding.py
```diff
@@ -0,0 +1,78 @@
+import pytest
+import canmatrix.formats
+import os.path
+
+def loadDbc():
+    here = os.path.dirname(os.path.realpath(__file__))
```
altendky: I put this in global scope so it's accessible to all.  Since you accepted my PR with pathlib2 you can also `pathlib2.Path(__file__).resolve().parent`.  (i missed the resolve part in my use)  Below you just `here/'test_frame_decoding.dbc'`.

src/canmatrix/canmatrix.py
ebroecker: yes, it is,
I could imagine that it's good to have a method which decodes just every signal regardless of the multiplexer status.
What your opinion as native speaker: what do you  think is better "decode/encode" or "pack/unpack" ?
Maybe unpack and _unpack_raw?

src/canmatrix/canmatrix.py
```diff
@@ -779,108 +869,172 @@ def updateReceiver(self):
...
-        :param bool decodeToStr: If True, try to get value representation as *string* ('Init' etc.)
         :return: OrderedDictionary
         """
+
+        rx_length = len(data)
+        if rx_length != self.size and report_error:
+            print(
+                'Received message 0x{self.id:08X} with length {rx_length}, expected {self.size}'.format(**locals()))
+        else:
+            if sys.version_info > (3,):
```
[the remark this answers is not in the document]
ebroecker: good point, gonna try to switch to bytearray if possible...

src/canmatrix/canmatrix.py
```diff
@@ -328,19 +333,72 @@ def calcMax(self):
 
         return self.offset + (rawMax * self.factor)
 
-    def bitstruct_format(self):
-        """Get the bit struct format for this signal.
+    def unpack_bitstring(self, length, is_float, is_signed, bits):
```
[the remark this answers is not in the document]
ebroecker: maybe some new module (tools? helpers?) for these functions? 
Also  "grouper" could be could in this new module?

src/canmatrix/canmatrix.py
```diff
@@ -328,19 +333,72 @@ def calcMax(self):
...
+        if is_float:
+            types = {
+                32: '>f',
+                64: '>d'
+            }
+
+            float_type = types.get(length)
+
+            if float_type is None:
+                raise Exception(
```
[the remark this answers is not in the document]
ebroecker: I took the second option. make the code smaller and more easy to understand

src/canmatrix/canmatrix.py
```diff
@@ -328,19 +333,72 @@ def calcMax(self):
...
+                    'float type only supports lengths in [{}]'.
+                        format(', '.join([str(t) for t in types.keys()]))
```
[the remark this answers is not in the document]
ebroecker: no exception anymore

src/canmatrix/canmatrix.py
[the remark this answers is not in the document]
ebroecker: I don't know why I put this into signal. I throwed it out to regular function again.
new module for these helpers/tools/utilities should be done in another pr.

src/canmatrix/canmatrix.py
```diff
@@ -779,108 +869,172 @@ def updateReceiver(self):
...
-        :rtype: bitstruct
         """
+
+
         data = dict() if data is None else data
 
         if self.is_complex_multiplexed:
             # TODO
-            pass
+            Exception("encoding complex multiplexed frames not yet implemented")
```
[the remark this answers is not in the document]
ebroecker: tried to create custom exceptions.

may be reworked. may be done in another pr

src/canmatrix/canmatrix.py
ebroecker: should be probably renamed. 
may be done in another pr.

src/canmatrix/tests/test_frame_decoding.py
ebroecker: feel free to change - I just did this simple solution when pathlib2 was not yet imported in project...

altendky: deferred for later resolution.  https://github.com/ebroecker/canmatrix/issues/236#issuecomment-435590542

PR #228, opened 2018-10-31
Tests for 'Define'

Funth0mas:
Simple tests for define. I don't know what is expected behaviour of `HEX`, but it seems suspicious that it behaves like `INT`. Can I check it somehow/somewhere? Is it based on DBC, or an another format?

### DIFF
--- a/src/canmatrix/tests/test_canmatrix.py
+++ b/src/canmatrix/tests/test_canmatrix.py
@@ -355,6 +355,64 @@ def test_encode_decode_frame():
     assert decoded_data == input_data
 
 
+# Define tests
+def test_define_set_default():
+    define = canmatrix.canmatrix.Define("")
+    define.setDefault("string")
+    assert define.defaultValue == "string"
+    define.setDefault('"quoted_string"')
+    assert define.defaultValue == "quoted_string"
+
+
+def test_define_update_enum_definition():
+    define = canmatrix.canmatrix.Define("")
+    define.type = "ENUM"
+    define.values = ["ready", "off"]
+    define.update()
+    assert define.definition == 'ENUM "ready","off"'
+
+
+def test_define_update_ingored_non_enum():
+    def_str = "INT 0 100"
+    define = canmatrix.canmatrix.Define(def_str)
+    define.update()
+    assert  define.definition == def_str
+
+
+def test_define_for_int():
+    define = canmatrix.canmatrix.Define("INT -5 10")
+    assert define.type == "INT"
+    assert define.min == -5
+    assert define.max == 10
+
+
+# def test_define_for_hex():
+#     define = canmatrix.canmatrix.Define("HEX ? ?")
+#     assert define.type == "HEX"
+#     assert define.min == ?
+#     assert define.max == ?
+
+
+def test_define_for_string():
+    define = canmatrix.canmatrix.Define("STRING")
+    assert define.type == "STRING"
+    assert define.min is None
+    assert define.max is None
+
+
+def test_define_for_enum():
+    define = canmatrix.canmatrix.Define('ENUM "red","green", blue')
+    assert define.type == "ENUM"
+    assert define.values == ["red", "green", "blue"]
+
+
+def test_define_for_float():
+    define = canmatrix.canmatrix.Define("FLOAT -2.2 111.11")
+    assert define.type == "FLOAT"
+    assert define.min == decimal.Decimal('-2.2')
+    assert define.max == decimal.Decimal('111.11')
+
+
 # CanId tests
 def test_canid_parse_values():
     can_id = canmatrix.canmatrix.CanId(0x01ABCD02)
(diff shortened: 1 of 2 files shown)


src/canmatrix/tests/test_canmatrix.py
```diff
@@ -355,6 +355,64 @@ def test_encode_decode_frame():
...
+
+def test_define_for_string():
+    define = canmatrix.canmatrix.Define("STRING")
+    assert define.type == "STRING"
+    assert define.min is None
+    assert define.max is None
+
+
+def test_define_for_enum():
+    define = canmatrix.canmatrix.Define('ENUM "red","green", blue')
```
altendky: Red and green are quoted but blue is not.  If the quotes provide any meaning let's exercise it.  Can enumeration names have spaces?  Special characters?  Comma inside quotes?

src/canmatrix/tests/test_canmatrix.py
```diff
@@ -355,6 +355,64 @@ def test_encode_decode_frame():
...
+    define.values = ["ready", "off"]
+    define.update()
+    assert define.definition == 'ENUM "ready","off"'
+
+
+def test_define_update_ingored_non_enum():
+    def_str = "INT 0 100"
+    define = canmatrix.canmatrix.Define(def_str)
+    define.update()
+    assert  define.definition == def_str
```
altendky:
```suggestion
    assert define.definition == def_str
```

src/canmatrix/tests/test_canmatrix.py
Funth0mas: The missing quotes and additional space are by intent to reflect the code in the function under the test. That means it again tests multiple things: quote removal, trim() used, all values parsed.
But it only reflects the code, I have no additional knowledge about it.

Funth0mas: You meant to split it to multiple tests, right?

altendky: Maybe multiple tests, but my point was to exercise some corner cases.  You obviously already covered some of them but I was suggesting more.  I'd either ~~comment~~ docstring (uh, hmm...) or make separate tests.  If the present code can handle something like `'ENUM "red", "green, blue, and cyan", magenta'` than let's have a test for the internal `,`s and ` `s inside the enumeration value.

ebroecker:
@Funth0mas 
do you plan to work on this PR (especially work on the findings from altendky)?

altendky:
I might just withdraw those comments, but let's see, since I don't really know defines.  It just seems bad to me to make the caller create a string and then parse that string back apart.  Why isn't it something like `canmatrix.canmatrix.create_enum_define('red', 'green', 'blue')`?  If we end up agreeing that the interface should change such that the idea of 'quoting values' doesn't apply, I'll withdraw the suggestion to test uses of those `"`s such as including commas or spaces in the value names.

altendky:
(just to be clear any such change should be in a separate PR or just ticketed for later decisions and implementation)

(also, if spaces and commas aren't supported even with a quoted name, then yeah, you wouldn't need to test it in that case either)

ebroecker:
well ... this is very near to _dbc_ once again :
defines are a kind of _datatypes_ for the additional _attributes_ in _dbc_. 
Any _attribute_ (frame, signal, general) in _dbc_ has to have a _datatype_. 
In dbc this is _datatype_ is declared by "DEF_" this is why I called it _define_.

My implementation just takes the string from dbc in canmatrix and saves it there.
This is probably very clumsy.
The other point is the name _define_ , not sure if it's what a native speaker would expect for a datatyp definition.

For this issue/test:
I'd leave it more or less as is. But we have to think about a clean redesign of these datatypes/defines in another pr. 
Especially for all not _dbc_ formats have to create their defines (if they want to be convertable to dbc) on their own now which is kind of black magic.

btw: there is a very old TODO for this issue:  https://github.com/ebroecker/canmatrix/blob/development/src/canmatrix/canmatrix.py#L28

Funth0mas:
I split the ENUM test to 2 or 3 functions, but no, complicated things like comma between quotes aren't supported anyway. @ebroecker do you know by chance if the HEX should really parse the min/max values like the INT, not using `int(value, 16)`? That's why I did not finished the test for HEX as I hesitate if it's correct. 
Or I can try to experiment with dbc editor, there is no hurry to merge this PR. It's test coverage only.

ebroecker:
@Funth0mas ,
I double checked, yes HEX and INT are the same in DBC, the only difference is the view.
grafik

results in:
```
BA_DEF_ SG_  "signal_attrib_definition_hex_0_2" HEX 0 581;
BA_DEF_ SG_  "signal_attrib_definition_int_0_2" INT 0 245;
```

src/canmatrix/tests/test_canmatrix.py
altendky: not supported so withdrawing suggestion

altendky:
More and more I'm buying into the approach of '`__init__()` should create an object, not parse and connect everything'.  `attrs` kinda ends up encouraging this since `__init__()` becomes slightly more awkward as `__attrs_post_init__()`.  Basically, objects should be able to be created directly, parsing etc that helps you make an object or insert it into other objects should be separate.  (I'm sure I regularly don't follow this myself but I'm trying)  (but yeah, that's not for this PR)

altendky:
kicking ci

PR #235, opened 2018-11-02
[WIP] Add .sym based encoding test

altendky:


ebroecker:
@altendky 
could you fix py2 issue?
After that, I don't see any problem with merging.

(it's really time to get rid of py2, but before I do this, I want to do a release...)

PR #241, opened 2018-11-07
Update dbc HEX define comments

altendky:
https://github.com/ebroecker/canmatrix/pull/228#issuecomment-435788100


altendky:
err, already addressed int #228...

Issue #237, opened 2018-11-02
Use loggers based on __name__


Funth0mas:
The canmatrix project tends to be more than a standalone converter only. I'd try to refactor loggers to use `getLogger(__name__)` instead of the root logger everywhere. The `setup_logger` would configure only top-level `canmatrix` logger. 
https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library.

altendky:
At least part of this got stalled in #164.  If we still like that we can revive it and get it merged then follow up with wherever else it needs tidied?

Funth0mas:
Yes I forgot to state that this is inspired by the mentionned PR, but it tried to fix single file only and the diff currently includes about three lines of code (among not being mergable currently).
I would do it globally (I don't know in how many files the logger is actually used). That is why I'm "asking" before working on it.

altendky:
I agree that `__name__` should generally be used. Perhaps in entry point functions it is proper to use `None` (explicitly or implicitly, no matter) and do the basic config or whatever?

ebroecker:
done?!

PR #244, opened 2018-11-09
Use module loggers.

Funth0mas:
As discussed in #237, use module loggers based on module names and only configure logger in main(). It is intended to replace+finish #164.

* every logger is child of main "canmatrix" logger
* if used as a standalone converter/comparator, the 'canmatrix' logger is configured to provide all messages
* if used as a library, the logger setting is up to the user. He can either configure its root logger or use default config by calling `canmatrix.log.setup_logger()`. Or leave default `logging` behavior.

If you agree with this change, it would be gentle to document the above points somewhere, but I don't know where.

The rows
```
if __name__ == '__main__':
    sys.exit(main())
```
in compare.py and convert.py don't work (at least due to import issues). It must be run using either examples/convert.py or the script generated by setuptool. I'll propose another PR to remove them, not to confuse anymore.

### DIFF
--- a/src/canmatrix/copy.py
+++ b/src/canmatrix/copy.py
@@ -21,9 +21,13 @@
 # DAMAGE.
 
 from __future__ import absolute_import
+import logging
 
 from .canmatrix import *
 from copy import deepcopy
+
+logger = logging.getLogger(__name__)
+
 
 def copyBU(buId, sourceDb, targetDb):
     """
--- a/src/canmatrix/log.py
+++ b/src/canmatrix/log.py
@@ -27,15 +27,18 @@ from __future__ import absolute_import
 import logging
 
 
-def setup_logger(name='canmatrix'):
-    """Setup the module logger for canmatrix module. Return logger instance for possible further setting."""
+def setup_logger():
+    """Setup the root logger. Return the logger instance for possible further setting and use.
+
+    To be used from CLI scripts only.
+    """
     formatter = logging.Formatter(
         fmt='%(levelname)s - %(module)s - %(message)s')
 
     handler = logging.StreamHandler()
     handler.setFormatter(formatter)
 
-    logger = logging.getLogger(name)
+    logger = logging.getLogger()
     logger.setLevel(logging.DEBUG)
     logger.addHandler(handler)
     return logger
--- a/src/canmatrix/log.py
+++ b/src/canmatrix/log.py
@@ -27,8 +27,8 @@ from __future__ import absolute_import
 import logging
 
 
-def setup_logger(name):
-    """Setup a project wide logger singleton"""
+def setup_logger(name='canmatrix'):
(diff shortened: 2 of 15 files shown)


src/canmatrix/log.py
```diff
@@ -27,8 +27,8 @@
 import logging
 
 
-def setup_logger(name):
-    """Setup a project wide logger singleton"""
+def setup_logger(name='canmatrix'):
```
altendky: I would have expected `name` to be either `None` or the `logging.getLogger()` call below to pass no parameters.  This would get the root logger which seems appropriate in the case of the cli scripts.  But, I certainly may be missing something.

src/canmatrix/xlsx.py
```diff
@@ -34,7 +34,7 @@
 import xlsxwriter
 from canmatrix.xls_common import *
 
-
+logger = logging.getLogger(__name__)
```
altendky: Do we need this if it's not being used?  I think it might be better not to preemptively add it.

src/canmatrix/log.py
```diff
@@ -27,8 +27,8 @@
...
+    """Setup the module logger for canmatrix module. Return logger instance for possible further setting."""
```
altendky:
```suggestion
    """Setup the logger.  By default this is the root logger. Return logger instance for possible further setting and use."""
```

If above is changed...

src/canmatrix/copy.py
```diff
@@ -21,10 +21,14 @@
 # DAMAGE.
 
 from __future__ import absolute_import
+import logging
 
 from .canmatrix import *
 from copy import deepcopy
 
+logger = logging.getLogger(__name__)
```
altendky: Maybe don't preemptively create this if unused.  Also noted below.

src/canmatrix/xlsx.py
Funth0mas: The logger was used in the xlsx.py file, but the instance came from `from .canmatrix import *` what is wrong. Now it uses own module logger.

src/canmatrix/copy.py
Funth0mas: Similar to xlsx, the copy module used wrong logger instance comming from `from .canmatrix import *`.

src/canmatrix/log.py
Funth0mas: The idea was to provide the library user with possibility to *only* configure 'canmatrix' logger tree without actually knowing the logger names. 
But the cli convert/compare scripts should probably call `setup_logger(None)`
I'll try to investigate more on common usage in libs.

src/canmatrix/xlsx.py
altendky: ack!  new item for the refactor list, no `*` imports.  :]

Funth0mas:
The `setup_logger()` only sets the root logger, meant to be used internally only. Added NullHandler as proposed in logging manual and used for example in requests lib.

Issue #226, opened 2018-10-29
Issue in converting arxml to dbc


Jayanthdj:
When I tried to convert .arxml to dbc file, I got this below error
![image](https://user-images.githubusercontent.com/33441532/47639616-c19a8f80-db87-11e8-9823-2c808d663a0a.png)

ebroecker:
Initvalue translation is buggy.

https://github.com/ebroecker/canmatrix/blob/master/canmatrix/arxml.py#L1140-L1145

hard coded lable to value translation fails if lable is something other than "true" or "false"

maybe this lookup is not bullet prove:
https://github.com/ebroecker/canmatrix/blob/master/canmatrix/arxml.py#L1051

Jayanthdj:
Hi,
After doing the modification you suggested, I am getting this below error
if initvalue is None or initvalue.__len__() == 0:
                initvalue = arGetXchildren(isignal, "INIT-VALUE/NUMERICAL-VALUE-SPECIFICATION/VALUE", arDict, ns) ##AR4.2
            elif initvalue is not None and initvalue.__len__() >= 1:
				initvalue = initvalue[0]

![image](https://user-images.githubusercontent.com/33441532/47707196-38e82600-dc51-11e8-907e-d74719205c10.png)

Jayanthdj:
I tried to convert arxml to dbc  with 'Canmatrix development' and got this below error
![image](https://user-images.githubusercontent.com/33441532/47987119-b950d000-e104-11e8-9e5d-5dda030b747c.png)

and with Canmatrix master I got this below error  :(
![image](https://user-images.githubusercontent.com/33441532/47987167-e2716080-e104-11e8-8187-1462bc9cb094.png)

ebroecker:
could you - maybe - provide an example also? PN?

Jayanthdj:
Hii, I sent

ebroecker:
Thanks,

not sure, your ARXML is correct:
```xml
<INIT-VALUE>
                <NUMERICAL-VALUE-SPECIFICATION>
                  <VALUE>On</VALUE>
```

in my unserstanding "On" is not realy numerical. But maybe this would be translated by the compuMethod '/CompuMethods/OnOff1_1'. But as far as I can see, this compuMethod is not included in the ARXML.

I'll provide a patch for fixing On and Off - but not sure if its the right way to do it...

ebroecker:
merged

Issue #236, opened 2018-11-02
Move converters to own module?


Funth0mas:
The converters are currently mixed with the core in the single module canmatrix. They even don't have any naming convence (like cm*.py). What about to move them to own module, say, converters? (...and later refactor them to classes with common abc interface...)

altendky:
I'm not big on the `cm` prefixes as we already have the canmatrix package (the outermost one).  It would be nice to never `canmatrix.canmatrix` as the second one really describes nothing useful.

As to interfaces, my impression from discussions in `#python` is that `zope.interface` is the thing to learn from and use, not stdlib (such as abc).

Anyways, you are talking about moving `canmatrix.dbc` to `canmatrix.converters.dbc`?  If so they are presently referred to as `formats` I think?

altendky:
If we're talking about an overall organizational review (I'd rather make an overall plan first then shift everything around, I'm sure some opinions will change as we discuss various points) then I'll add that we might need a misc/utils or whatever module for things like `grouper()` (unless we use one of the libraries that provides tools like this).

https://github.com/ebroecker/canmatrix/pull/224/files#diff-c1624d42e3750ac47cb2bbd3c178a5d0R536

Funth0mas:
Bad name, of course I thought `formats`.
Let cancel this ticket and make it somehow better...later... But I'm afraid I won't help a lot with it :-(

Funth0mas:
Concerning the prefix, only now I get the `cm` stands for 'canmatrix', not 'converter'. Yes, I'm not analphabet, I see the converter is not comverter :-) . Does not matter.

altendky:
We can rename the ticket or make a new one, no matter to me.  I see a few things at the moment, let's build the list if there are more.  If we are expecting to get more tests in place soone than later, maybe we continue that first before any reorganization.

* Add `canmatrix.formats` package
* `canmatrix.canmatrix` module should be renamed
* Possibly expose a particular few things via `canmatrix.some_thing()`

altendky:
* This isn't urgent for me so waiting until after a release seems fine.
* I expected to give more notice than that for dropping py2.  I don't know how we go about making sure people notice but I would say we at least have a release which notifies of imminent loss of py2 support before releasing a version that doesn't support py2 (or makes no promises of support for it).  I only brought it up because I don't want to be continuing to support it in 2025 simply because we didn't get around to warning anyone it was going away.
*
[...]

PR #247, opened 2018-11-19
[WIP] fix for #226

ebroecker:
fix for missing "putSignalIntoFrame"

### DIFF
--- a/src/canmatrix/arxml.py
+++ b/src/canmatrix/arxml.py
@@ -1204,14 +1204,14 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
                     newSig.addValues(1, "TRUE")
                     newSig.addValues(0, "FALSE")
 
-            def guessValue(textValue):
+            def guess_value(textValue):
                 """
-                returns a numerical value for common strings.
+                returns a string value for common strings.
                 method is far from complete but helping with odd arxmls
                 :param textValue: value in text like "true"
                 :return: string for value like "1"
                 """
-                textValue = textValue.lower()
+                textValue = textValue.casefold()
                 if textValue in ["false", "off"]:
                     return "0"
                 elif textValue in ["true", "on"]:
@@ -1219,7 +1219,7 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
                 return textValue
 
             if initvalue is not None and initvalue.text is not None:
-                initvalue.text = guessValue(initvalue.text)
+                initvalue.text = guess_value(initvalue.text)
                 newSig._initValue = int(initvalue.text)
                 newSig.addAttribute("GenSigStartValue", str(newSig._initValue))
             else:
--- a/src/canmatrix/arxml.py
+++ b/src/canmatrix/arxml.py
@@ -1200,12 +1200,22 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
                     newSig.addValues(1, "TRUE")
                     newSig.addValues(0, "FALSE")
 
+            def guessValue(textValue):
+                """
+                returns a numerical value for common strings.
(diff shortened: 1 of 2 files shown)


src/canmatrix/arxml.py
```diff
@@ -1200,12 +1200,22 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
                     newSig.addValues(1, "TRUE")
                     newSig.addValues(0, "FALSE")
 
+            def guessValue(textValue):
```
altendky: I don't know where we are on this 'officially' but do we and `lowerCamel` or `snake_case`?  I'll add this to the refactor ticket to consider.

https://github.com/ebroecker/canmatrix/issues/236#issuecomment-435590542

src/canmatrix/arxml.py
```diff
@@ -1200,12 +1200,22 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
...
+                """
+                returns a numerical value for common strings.
```
altendky:
```suggestion
                returns a numerical string for common strings.
```
At least that's what it does presently.

src/canmatrix/arxml.py
```diff
@@ -1200,12 +1200,22 @@ def getSignals(signalarray, Bo, xmlRoot, ns, multiplexId, float_factory):
...
+                method is far from complete but helping with odd arxmls
+                :param textValue: value in text like "true"
+                :return: string for value like "1"
+                """
+                textValue = textValue.lower()
+                if textValue in ["false", "off"]:
+                    return "0"
+                elif textValue in ["true", "on"]:
+                    return "1"
```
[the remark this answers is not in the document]
ebroecker: good idea.

src/canmatrix/arxml.py
ebroecker: ... so you preferre `snake_case` . I tried to create new features in `snake_case` but this time, I missed my intent

altendky: Should I change the linked ticket to just state it will be 'snake_case'?  I suppose that we could instead be picking an existing coding standard/checker (black, or something configurable) and that would cover many of these questions.

src/canmatrix/arxml.py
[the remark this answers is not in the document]
altendky: It has the cost of being less flexible for handling situations like if you only want some things to be case insensitive etc.  But either way, it's small now and could be easily readjusted either direction in the future as needed.

altendky:
@ebroecker, should I be a stickler for coverage?  If so, go add tests...  if not, merge away.  :]

ebroecker:
> 
> 
> @ebroecker, should I be a stickler for coverage? If so, go add tests... if not, merge away. :]

@altendky 
coverage for arxml is a hard job. 
I'd have to create a 'dummy-arxml' first. 
I'll do this in another PR

altendky:
While we should have integration tests, it would be nice for more and more test coverage to come from tests closer to the unit level.  For example, for this reason and others, `guess_value()` should be defined in global scope and can then be directly tested.  Also, unless there's going to be a complete replacement of the parsing code (we at least vaguely discussed parsley or other parsing libraries) it should be broken down into smaller functions that can be called individually.  For example, without having checked the details we ought to be able to call a function that parses a single option from a line in a sym file, and another function to parse a single line (perhaps a different function for lines in different places).

So, easy enough to get coverage on `guess_value()` here.  Consider the couple tests and 'files' in [`canmatrix.tests.sym`](https://github.com/ebroecker/canmatrix/blob/9dfd08efe39980be459a7377e4f28da9b07efc7d/src/canmatrix/tests/test_sym.py).  My intent is that the file only contains the minimum necessary to execute the test.  The point is to make it clear what is involved in the feature being tested and also that if we break something else, this test is less likely to fail.

Issue #248, opened 2018-11-20
.sym file long names are truncated if they contain a colon


toddsnider-epc:
In the .sym file below, the long name "Access Level : Password" is truncated to "Access Level."
```
FormatVersion=5.0 // Do not edit this line!
Title="a file"

{SEND}

[Parameter]
DLC=8
Var=Password unsigned 16,16 /ln:"Access Level : Password"
```

PR #249, opened 2018-11-20
Prevent truncation of .sym file long names

toddsnider-epc:
ebroecker/canmatrix#248

### DIFF
--- a/src/canmatrix/tests/test_sym.py
+++ b/src/canmatrix/tests/test_sym.py
@@ -48,7 +48,7 @@ def test_colliding_mux_values():
     assert 'SecondMux' in error_string
 
 
-def test_parse_longname():
+def test_parse_longname_with_colon():
     f = io.BytesIO(
         textwrap.dedent(
             '''\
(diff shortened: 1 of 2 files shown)


src/canmatrix/tests/test_sym.py
```diff
@@ -46,3 +46,27 @@ def test_colliding_mux_values():
 
     assert 'FirstMux' in error_string
     assert 'SecondMux' in error_string
+
+
+def test_parse_longname():
```
altendky:
```suggestion
def test_parse_longname_with_colon():
```
Delineates it from a 'generic' long name parsing test.

altendky:
Heh, the `if ln:` actually causes coverage loss because it's only exercised one way.

https://codecov.io/gh/ebroecker/canmatrix/pull/249/diff?src=pr&el=tree#diff-c3JjL2Nhbm1hdHJpeC90ZXN0cy90ZXN0X3N5bS5weQ==

@ebroecker, it seems a bit weird/bad that codecov is responding with a pass for code that is added but not covered properly.  Maybe it's only because it is a conditional and it is half covered?  (the True case, but not the False case).  Anyways, I don't offhand even see any settings on the codecov site for this so maybe we just keep an eye on it so we know how it works.

PR #253, opened 2018-11-21
Iss243

### LINKED ISSUE #243: Request Configure Min-max value when we config Value Decription ( Value Table) in excel file for convert Excel to dbc
Hi,

Normally, in file excel, we can configure the Min, Max value in the Value column ( min..max).
But when we configure for Value description. The Max value will be the largest value of Signal Length [Bit] (4 bit => max = 15 ).
Even I use the signal.max column. The value is still the largest value of Signal Length [Bit]. 

Ex: My signal have range from 0 - 250. And 255 will be invalid value. 
The converted dbc is always have max value = 255 even I set signal.max = 250 like below

The value of my column:
Signal   Name = Signal_A
Signal Function = Speed 
Signal Length [Bit] = 8
Signal Default = blank
Signal Not Available = blank
Byteorder = m
Node_a = s
Value = 255
Name / Phys. Range = Invalid
Function / Increment Unit = 0.4 m/s
signal.offset = 0
signal.max = 250

ebroecker:
should fix  #243


src/canmatrix/xls.py
```diff
@@ -309,6 +309,28 @@ def dump(db, file, **options):
     # save file
     workbook.save(file)
 
+############################ load ###############################
+
+def parse_value_name_collumn(value_name, value, signal_size, float_factory):
```
altendky:
```suggestion
def parse_value_name_column(value_name, value, signal_size, float_factory):
```

src/canmatrix/xls.py
```diff
@@ -309,6 +309,28 @@ def dump(db, file, **options):
...
+    mini = maxi = offset = None
+    value_table = dict()
+    if ".." in value_name:
+        (mini, maxi) = value_name.strip().split("..", 2)
+        mini = float_factory(mini)
+        maxi = float_factory(maxi)
+        offset = mini
+
+    elif value_name.__len__() > 0:
```
altendky:
```suggestion
    elif len(value_name) > 0:
```

src/canmatrix/xls.py
```diff
@@ -309,6 +309,28 @@ def dump(db, file, **options):
...
+        if value.strip().__len__() > 0:
```
altendky:
```suggestion
        if len(value.strip()) > 0:
```

altendky:
@ebroecker, let me know when this should get a real review.

ebroecker:
@altendky Yes, If you want to and have 'complaints' - else I'd merge just 'as is'.

src/canmatrix/xls.py
altendky: Search and replace `collumn` everywhere with `column`.

https://en.wiktionary.org/wiki/column

requirements.tox.txt
```diff
@@ -3,3 +3,5 @@ py==1.5.4
 six==1.11.0
 tox==3.2.1
 virtualenv==16.0.0
+xlwt==1.3.0
+xlrd==1.1.0
```
altendky: These aren't required to install `tox` itself.  Only in the requirements files above which tox uses to install in the envs it creates.  And yes, this is a mess.  I don't like how it is (my fault) nor do I have an overall good solution to it presently.

src/canmatrix/xls.py
```diff
@@ -309,6 +303,28 @@ def dump(db, file, **options):
     # save file
     workbook.save(file)
 
+############################ load ###############################
+
+def parse_value_name_collumn(value_name, value, signal_size, float_factory):
+    mini = maxi = offset = None
+    value_table = dict()
+    if ".." in value_name:
+        (mini, maxi) = value_name.strip().split("..", 2)
```
altendky:
```suggestion
        (mini, maxi) = value_name.strip().split("..")
```

The second argument is the maximum number of splits, not the maximum number of elements.  So `2` would allow for three elements which would fail here.  Since we want to fail on too many `..`s anyways, we can just do it here with the assignment.  Though a more explicit error could be made if we split and stored the result and checked its length.  `*shrug*`

src/canmatrix/xls.py
```diff
@@ -309,6 +303,28 @@ def dump(db, file, **options):
...
+
+    elif len(value_name) > 0:
+        if len(value.strip()) > 0:
+            # Value Table
+            value = int(float(value))
+            value_table[value] = value_name
+        maxi = pow(2, signal_size) - 1
+        maxi = float_factory(maxi)
+        mini = 0
+        offset = 1
```
altendky: This seems odd having a default offset of 1, but I don't really know what this is doing so...  :|  I'll highlight it like this but if it's proper, just move on and ignore this comment.

src/canmatrix/xls.py
```diff
@@ -539,23 +555,17 @@ def load(file, **options):
...
-                value = int(float(value))
-                newSig.addValues(value, valueName)
-            maxi = pow(2, signalLength) - 1
-            newSig.max = float_factory(maxi)
-        else:
-            newSig.offset = 0
+        if newSig.offset is None:
+            newSig.offset = offset
+        if value_table is not None:
+            for val in value_table:
```
altendky:
```suggestion
            for value, name in value_table.items():
```

src/canmatrix/xls.py
```diff
@@ -539,23 +555,17 @@ def load(file, **options):
...
+                newSig.addValues(val, value_table[val])
```
altendky:
```suggestion
                newSig.addValues(value, name)
```

src/canmatrix/xls.py
ebroecker: thank you for this hint, my spelling is not the best, ...

src/canmatrix/xls.py
ebroecker: you are right here, I thought about factor when I initialized offset with 1.

src/canmatrix/tests/test_xls.py
```diff
@@ -0,0 +1,19 @@
+import canmatrix.xls
+import decimal
+
+
+def test_parse_value_name_collumn():
```
altendky: Five more instances of `collumn` to fix.

ebroecker: oh no

PR #254, opened 2018-11-21
rename findNotUsedBits in get_frame_layout

### LINKED ISSUE #227: findNotUsedBits() reports the signals index+1, not the index
https://github.com/ebroecker/canmatrix/blob/df1c7b78da60701bfecab01df513b2c3f0903eef/src/canmatrix/canmatrix.py#L711-L745

1) I would think this should be filling out indexes that can actually be used to index the signal list, or just filling out the actual signal.  I tend towards the latter unless there's a specific reason anyone needs to care about the index.
2) Use `for index, signal in enumerate(self.signals):` if we are going to stick with the indexes.
3) Regardless, `None` is a much more clear indicator of nothing than 0.
4) In my packing code (being ported in in #224) there is allowance for overlapping signals.  To retain that each element should be a list, not a single value.  If this is used then an empty list would indicate nothing as opposed to None.

ebroecker:
should once fix #227

let get_frame_layout create array for each bit containing signals using this bit
fix createDummySignals to use  get_frame_layout
rename createDummySignals to create_dummy_signals


src/canmatrix/canmatrix.py
```diff
@@ -815,67 +815,55 @@ def calcDLC(self):
...
-                if sig.is_little_endian:
-                    bitfieldLe[bit] = i
-                else:
-                    bitfieldBe[bit] = i
-
-        for i in range(0,8):
-            for j in range(0,8):
-                bitfield[i*8+j] = bitfieldLe[i*8+(7-j)]
+        little_bits = [[] for _dummy in range(0, (self.size * 8))]
+        big_bits = [[] for _dummy in range(0, (self.size * 8))]
```
```suggestion
        big_bits = [[] for _dummy in range((self.size * 8))]
```

src/canmatrix/canmatrix.py
```diff
@@ -815,67 +815,55 @@ def calcDLC(self):
...
 
-        for i in range(0,8):
-            for j in range(0,8):
-                if bitfield[i*8+j] == 0:
-                    bitfield[i*8+j] = bitfieldBe[i*8+j]
+        little_bits = reversed(tuple(grouper(little_bits, 8)))
+        little_bits = tuple(chain(*little_bits))
 
+        returnList = list()
+        for i in range(0,self.size * 8):
```
```suggestion
        for i in range(self.size * 8):
```

But overall this could be:

```python
returnList = [
    little + big
    for little, big in zip(little_bits, big_bits)
]
```

src/canmatrix/canmatrix.py
```diff
@@ -815,67 +815,55 @@ def calcDLC(self):
...
-            for bit in range(sig.getStartbit(),  sig.getStartbit() + int(sig.size)):
-                if sig.is_little_endian:
-                    bitfieldLe[bit] = i
-                else:
-                    bitfieldBe[bit] = i
-
-        for i in range(0,8):
-            for j in range(0,8):
-                bitfield[i*8+j] = bitfieldLe[i*8+(7-j)]
+        little_bits = [[] for _dummy in range(0, (self.size * 8))]
```
```suggestion
        little_bits = [[] for _dummy in range((self.size * 8))]
```

Doesn't do anything here.

src/canmatrix/canmatrix.py
```diff
@@ -815,67 +815,55 @@ def calcDLC(self):
...
         """
-        bitfield = self.findNotUsedBits()
-        # for i in range(0,8):
-        #    print (bitfield[(i)*8:(i+1)*8])
+        bitfield = self.get_frame_layout()
         startBit = -1
         sigCount = 0
-        for i in range(0,64):
-            if bitfield[i] == 0 and startBit == -1:
+        for i in range(0,self.size * 8):
```
```suggestion
        for i in range(self.size * 8):
```

How about using `len(bitfield)` throughout instead of `self.size * 8`?

And could we do something like below?  We do use the index but at least we can drop the `range()` and indexing

```python3
for index, bit_signals in enumerate(bitfield):
```

ebroecker:
@altendky 
any further suggestions or can I just merge this one?

Thanks again

src/canmatrix/canmatrix.py
altendky: @ebroecker, while we do need the index in this case it's nice to at least tie the iteration to some existing iterable.  So this is about `for index, bit_signals in enumerate(bitfield):` and also using `len(bitfield)` everywhere in this function where there is presently `self.size * 8`.

src/canmatrix/canmatrix.py
altendky: @ebroecker, using indexes or `range()` when they are not needed is a bit of a pet peeve of mine.

src/canmatrix/canmatrix.py
ebroecker: OK, usage of len(bitfield) is nicer.
The other idea - using  `enumerate(bitfield)` - is this more easy to understand than using the index? If so, I'll change

altendky: Most of the time we don't care about indexes so using
```python
for i in range(len(things)):
    do_something_with(things[i])
```
is much more simply written as
```
for thing in things:
    do_something_with(thing)
```
On the occasion that we do care about indexes that is easily expanded to
```
for i, thing in enumerate(things):
    do_something_with(i, thing)
```
which still leaves us iterating over the stuff of interest but also getting an index for each item.  I suppose there's room for opinion here but I much prefer using `enumerate()` to add an index to iteration than to not even iterate over the thing i want each element of.

src/canmatrix/canmatrix.py
ebroecker: .... I can't hide my c language experiences....  
I'm learning every day more thinking 'pythonic' working with you :-) 
So thank you very much for your hints

src/canmatrix/canmatrix.py
ebroecker: Ok, I think I understand.
But this time, we only need the index and do not need the content of bitfield. 
So this time it seems to me more confusing to me to iterate while I only need the index, or do I miss something?

altendky: Aren't there at least a couple accesses to `bitfield[i]`?

image

src/canmatrix/canmatrix.py
```diff
@@ -809,70 +809,59 @@ def calcDLC(self):
...
-            for j in range(0,8):
-                bitfield[i*8+j] = bitfieldLe[i*8+(7-j)]
+        little_bits = [[] for _dummy in range((self.size * 8))]
+        big_bits = [[] for _dummy in range((self.size * 8))]
+        for signal in self.signals:
+            if signal.is_little_endian:
+                least = len(little_bits) - signal.startBit
+                most = least - signal.size
+                for i in range(most, least):
+                    little_bits[i].append(signal)
```
```python
for little_bit_signals in little_bits[most:least]:
    little_bit_signals.append(signal)
```

Similarly below for the big bits.  I think this is the same...

src/canmatrix/canmatrix.py
ebroecker: OK, got it, convinced.

src/canmatrix/canmatrix.py
ebroecker: convinced.

Issue #255, opened 2018-11-21
Signal min/max defaults don't respect calc_min_for_none and calc_max_for_none


toddsnider-epc:
The following test fails:
```
def test_signal_min_unspecified_respects_calc_for_min_none_false():
    signal = canmatrix.Signal(calc_min_for_none=False)
    assert signal.min is None
```

altendky:
This is one of the most features of a PR.  I think they should really have two layers of changes.  One that introduced tests that fail. Another that builds on that too fix the failing tests.  Then you wouldn't have to explain which one failed...  Anyways

PR #256, opened 2018-11-21
Make Signal min/max defaults respect calc_xxx_for_none flags.

toddsnider-epc:
ebroecker/canmatrix#255


altendky:
Todd and I talked about this and the whole situation is fiddly, but to fix the regression...  ok. 
 Longer term we should figure out how to do this more cleanly.  We may need a signal builder function, or just require an explicit calc limits call if that's wanted.  Or limits could be properties and calculated on the fly if needed and so configured.

If you want to read opinions related to this situation (that of working with a partially initialized attr class) you can read up below.

https://github.com/python-attrs/attrs/issues/240
https://github.com/python-attrs/attrs/pull/404

PR #262, opened 2018-11-25
[WIP] rename copy module due to namespace collision with global namespace

ebroecker:


src/canmatrix/__init__.py
```diff
@@ -9,7 +9,7 @@
 import canmatrix.formats as formats
 import canmatrix.cancluster as cancluster
 import canmatrix.convert as convert
-import canmatrix.copy as copy
+import canmatrix.cmcopy as cmcopy
```
altendky: If anything is needed I believe `import canmatrix.copy` is sufficient.  But really if someone wants a submodule they should import it themselves, not expect importing it's parent to import it.

src/canmatrix/canmatrix.py
```diff
@@ -44,11 +41,12 @@
     from itertools import zip_longest as zip_longest
 except ImportError:
     from itertools import izip_longest as zip_longest
+
 from itertools import chain
 import struct
 
 from past.builtins import basestring
-import copy 
+from . import cmcopy
```
altendky: Just `import canmatrix.copy` I would think?  `cmcopy` is used once.

src/canmatrix/convert.py
```diff
@@ -31,7 +31,7 @@
 sys.path.append('..')
 import canmatrix.formats
 import canmatrix.canmatrix as cm
-import canmatrix.copy as cmcp
+import canmatrix.cmcopy as cmcp
```
altendky: We use `cmcp` five times, can we really not write `canmatrix.copy` instead?

ebroecker:
@wow - I could have bet that I tried this before, but seems it works now - the cleanest approach for the namespace issue...

PR #264, opened 2018-11-29
Tidy: pep8 fix spaces in attributes

Funth0mas:
Remove spaces around '=' signs in attr.ib according to pep8 to decrease number of warning (currently only) in the IDE.
No code change.


altendky:
Kick CI, silly pypy again.

PR #265, opened 2018-11-30
[WIP] Iss253 - Not yet able to merge!

ebroecker:
mySplit2 is not really what I want to have as solution, searching for better solution...

But either `shlex.split` nor `csv.reader` are able to split `'"Val1",","'` correct.
Expected value would be `["Val1",","]` or   `['"Val"','","']`.
[try online]( http://tpcg.io/Y8dUFi )

@altendky : Do you have an idea?


ebroecker:
maybe csv is a way what works:
http://tpcg.io/3T5ZAR

Issue #275, opened 2019-01-14
Export value range(min/max) of SG signals to JSON


nodermatt:
First of all, thanks for the fantastic tool @ebroecker ! It was extremely helpful for us when we developed an App that needs to handle many different dbc files for various vehicle types.

I've got one question regarding the json extraction option (--jsonExportAll ): Is it possible to extract the valid value range of a CAN message from dbc to JSON also?

We noticed that when converting the below CAN message from dbc we're missing the min and max value defined in square brackets between the "unit" and "factor/offset". There is an issue [#126](https://github.com/ebroecker/canmatrix/issues/126 )about min/max/default but I wasn't sure if it is related to my question.

```
BO_ 1568 C2C_BCM_TimeSync: 6 C2C
 SG_ C2C_BCM_noUnixTimestamp : 0|40@1+ (1,0) [0|1099511627775] "s"  BCM
 SG_ C2C_BCM_UtcOffsetTimezone : 40|7@1+ (15,-720) [-720|840] "minutes"  BCM
 SG_ C2C_BCM_DST : 47|1@1+ (1,0) [0|1] ""  BCM
```

I thought I might be able to extend the code in cmjson.py [here] https://github.com/ebroecker/canmatrix/blob/b11574484ce03723a327de76f6b7a35c2aeaa250/src/canmatrix/cmjson.py#L134) as there is a min and max field for the "SG_" decoded signals which I found in dbc.py https://github.com/ebroecker/canmatrix/blob/b11574484ce03723a327de76f6b7a35c2aeaa250/src/canmatrix/dbc.py#L513

Is there a particular reason why the min and max fields aren't exported by default?

```
{
      },
      "comment": "",
      "id": 1568,
      "is_extended_frame": false,
      "length": 6,
      "name": "C2C_BCM_TimeSync",
      "signals": [
          "bit_length": 40,
          "comment": "UTC Zeit als Unix Zeitstempel (Sekunden seit 1.1.1970)",
          "factor": "1",
          "is_big_endian": false,
          "is_float": false,
          "is_signed": false,
          "name": "C2C_BCM_noUnixTimestamp",
          "offset": "0",
          "start_bit": 0,
          "unit": "s",
          "values": {}
        },
          "bit_length": 7,
          "comment": "Offset der Lokalen Zeitzone zu UTC in Minuten (ohne Sommerzeit)",
          "factor": "15",
          "is_big_endian": false,
          "is_float": false,
          "is_signed": false,
[...]

PR #276, opened 2019-01-15
add min and max to jsonAll export #275

ebroecker:
fixes 
* min and max for json import/export
* add tests for this
* remove import * from canmatrix

### DIFF
--- a/src/canmatrix/cmjson.py
+++ b/src/canmatrix/cmjson.py
@@ -25,11 +25,11 @@
 # (https://github.com/ericevenchick/CANard)
 
 from builtins import *
-from .canmatrix import *
+import canmatrix
 import codecs
 import json
 import sys
-
+import decimal
 
 
 extension = 'json'
@@ -178,7 +178,7 @@ def dump(db, f, **options):
 
 
 def load(f, **options):
-    db = CanMatrix()
+    db = canmatrix.CanMatrix()
 
     if (sys.version_info > (3, 0)):
         import io
@@ -189,7 +189,7 @@ def load(f, **options):
     if "messages" in jsonData:
         for frame in jsonData["messages"]:
             #            newframe = Frame(frame["id"],frame["name"],8,None)
-            newframe = Frame(frame["name"],
+            newframe = canmatrix.Frame(frame["name"],
                              id=frame["id"],
                              size=8)
             if "length" in frame:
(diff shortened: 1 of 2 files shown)


src/canmatrix/cmjson.py
```diff
@@ -211,13 +213,20 @@ def load(f, **options):
...
+                newsignal = canmatrix.Signal(signal["name"],
                                    startBit=signal["start_bit"],
                                    size=signal["bit_length"],
                                    is_little_endian=is_little_endian,
                                    is_signed=is_signed,
                                    factor=signal["factor"],
                                    offset=signal["offset"])
+
+                if "min" in signal and signal["min"]:
+                    newsignal.min = decimal.Decimal(signal["min"])
```
ebroecker: @altenky 
I hast to add "decimal.Decimal"-casting here, 
not sure why ` min = attr.ib(` from signal-class did not work.
What do I wrong here?

altendky:
```suggestion
                    newsignal.min = newsignal.float_factory(signal["min"])
```

The converter for an `attr.ib()` is only applied when set through the constructor.  We could make `@property`s which apply the converter on assignment as well but, I dunno...  I do have a thing which does that automatically.  But, if we want that it should certainly be it's own effort.

Do beware when writing json as to whether you write floats or strings.  The former loses the benefit of using `decimal.Decimal()`.

Sorry I missed the review request...

src/canmatrix/cmjson.py
```diff
@@ -199,32 +200,42 @@ def load(f, **options):
                 newframe.extended = 0
 
             for signal in frame["signals"]:
-                if "is_big_endian" in signal and signal["is_big_endian"]:
+                if signal.get("is_big_endian", False):
```
altendky:
```suggestion
                is_little_endian = signal.get("is_big_endian", False)
```

While we're refactoring...  :]  Do note that these are both a little different than before, though maybe more what we want, maybe.  In the left-side code the result was `True` if the value wasn't specified.  Not sure if you want to default to little or big endian.

src/canmatrix/cmjson.py
```diff
@@ -199,32 +200,42 @@ def load(f, **options):
...
                                    startBit=signal["start_bit"],
                                    size=signal["bit_length"],
                                    is_little_endian=is_little_endian,
                                    is_signed=is_signed,
+                                   is_float = is_float,
                                    factor=signal["factor"],
                                    offset=signal["offset"])
-                if "unit" in signal and signal["unit"]:
+
+                if signal.get("min", False):
```
altendky:
```suggestion
                if signal.get("min") is not None:
```

What about a `0` minimum?  It'll be falsey...

PR #296, opened 2019-02-23
[WIP] introduction of Arbitration_id class

### LINKED ISSUE #144: handling of extended CAN id
Hello,

I have a DBC file that contains the following definition:

```
BO_ 2222222222 MSG_4746B8E: 6 MSG
```

If is load this DBC and try to get the message using the extended ID, the frame is not found.

Should the user handle the extended ids?

ebroecker:
should fix  #144


ebroecker:
@altendky 
could you have a look on the arbitration_id class?

It's not necessary to look on every single source line which had to be changed...

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
 
     return bitstring
 
+@attr.s(cmp=False)
+class Arbitration_Id(object):
```
```suggestion
class ArbitrationId(object):
```

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
...
+    standard_id_mask = ((1 << 11) - 1)
+    extended_id_mask = ((1 << 29) - 1)
+    compound_extended_mask = (1 << 31)
+
+    id = attr.ib(default=None)
+    extended = attr.ib(default=None)
```
```suggestion
    extended = attr.ib(default=True)
```

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
...
+
+    def __attrs_post_init__(self):
+        if self.extended is None or self.extended:
```
```suggestion
        if self.extended:
```

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
...
+            extended=(i & cls.compound_extended_mask) != 0,
+        )
+
+    def to_compound_integer(self):
+        if self.extended:
+            return self.id | self.compound_extended_mask
+        else:
+            return self.id
+
+    def __eq__(self, other):
```
altendky: I think we should be able to drop this after the above changes?  The default attrs-provided `__eq__` should work and if it doesn't we should probably try to look for a way to make it.

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
...
+@attr.s(cmp=False)
+class Arbitration_Id(object):
+    standard_id_mask = ((1 << 11) - 1)
+    extended_id_mask = ((1 << 29) - 1)
+    compound_extended_mask = (1 << 31)
+
+    id = attr.ib(default=None)
+    extended = attr.ib(default=None)
+
+    def __attrs_post_init__(self):
```
altendky: I'm tending away from writing `__attrs_post_init__`s anymore.  I think this could be just a helper `@classmethod` that the `from_compound_integer()` calls.  We should be _able_ to construct any class we want rather than having to create it and then modify to get to any arbitrary set of attributes.

src/canmatrix/canmatrix.py
```diff
@@ -551,6 +552,48 @@ def pack_bitstring(length, is_float, value, signed):
...
+        if self.extended is None or self.extended:
+            mask = self.extended_id_mask
+        else:
+            mask = self.standard_id_mask
+
+        if self.id != self.id & mask:
+            raise ArbitrationIdOutOfRange('ID out of range')
+
+    @classmethod
+    def from_compound_integer(cls, i):
```
```suggestion
    def from_compound_integer(cls, i, extended=True):
```

And associated changes inside the function.

src/canmatrix/canmatrix.py
```diff
@@ -574,10 +617,10 @@ class Frame(object):
     """
 
     name = attr.ib(default="")
-    id = attr.ib(type=int, default=0)
+    arbitration_id = attr.ib(converter=Arbitration_Id.from_compound_integer, default=0)
```
altendky: Do we need to be so verbose re: `arbitration_id` over `id`?  It might be better, but I'm not sure there's likely much confusion with just `id`.  Either way.

src/canmatrix/canmatrix.py
```diff
@@ -574,10 +617,10 @@ class Frame(object):
...
     size = attr.ib(default=0)
     transmitters = attr.ib(type=list, factory=list)
-    extended = attr.ib(type=bool, default=False)
+    # extended = attr.ib(type=bool, default=False)
```
altendky: Hmm, So we're changing from `default=False` to `default=True` effectively above.  I'm not sure which it should be, or how much it matters.

Do delete this before it gets left hanging around.

src/canmatrix/canmatrix.py
Funth0mas: Attention, `id` is a python keyword. It would lead to "shadowing" warning.

altendky: As it presumably has been already.  I can't say I like it but it only shadows in the class definition itself.  Once you are outside the class definition there's no more shadowing since it's `self.id`.  Since you mostly shouldn't use `id()` at all anyways...  I think I'm ok with this.  It's a hazard of mixing declarative with imperative like attrs does in this style usage.

src/canmatrix/canmatrix.py
ebroecker: there was the requirement to get both arbitration ids - the extended and the basic one - with the "by_id" method. Therefore the "None" was. If you set extended to "None" and compare two arbitration ids you'll get a true for both - the extended and the basic version.

src/canmatrix/canmatrix.py
ebroecker: ... the `compound_integer` already has the `extended`information in bit `31` - so no need for the extended attribute. Or do I miss something?

src/canmatrix/canmatrix.py
altendky: Hmm...  maybe `by_id()` should be `return itertools.chain.from_iterable(by_id(id, extended) for type in (False, True))` or less niftily `return by_id(id, False) + by_id(id, True)`. (aside from the recursion in both cases...).  I'm really not sure.  I'm just taking to simpler classes at this point.  :|  Do what you want and we'll see how we feel about it later.

src/canmatrix/canmatrix.py
altendky: :[ oh right, it's _compound_...  I'm really not on my game today, sorry.

ebroecker:
OK, I'll merge - we can work on it later again...

PR #297, opened 2019-02-23
Try using 3.6 to create venv on Windows

altendky:


altendky:
@ebroecker, well, you at least get everything but `pypy` back.  It could be marked as allowed to fail on AppVeyor and left for another day...  or, not.

ebroecker:
Thanks a lot!
Pypy3 even did not run before. (PR #195).
So it's for now OK  for me to ignore pypy in the windows build till er get an idea whats going wrong.

PR #298, opened 2019-02-26
Make classes upper case again

Funth0mas:
When merging the 'typing' branch I realized that some classes was unwillingly refactored to snake case.


ebroecker:
... what do you think, should classes be snake_case or camelCase?
I don't know.
We only discussed lossely in #236 to switch to snake_case.

In all cases, you are right, its not consistent as is now.

altendky:
`UpperCamel` is what pep8 suggests for normal situations.

https://www.python.org/dev/peps/pep-0008/#class-names

PR #301, opened 2019-02-26
[WIP] move formats to submodule (#236)

ebroecker:
is this the way to restructure the formats to an own module?


altendky:
I think so.

PR #303, opened 2019-02-27
Resolve inspection issues in `formats` module.

Funth0mas:
- snake case
- newlines
- spaces


Funth0mas:
ugh this is in conflict with ebroeckers' branch, cancelling it.

PR #304, opened 2019-02-27
Tidy and annotate CanCluster

Funth0mas:
- Upper case class name
- Snake case variables
- Define used instance variables in `__init__`
- Annotate

### DIFF
--- a/src/canmatrix/cancluster.py
+++ b/src/canmatrix/cancluster.py
@@ -1,74 +1,84 @@
-class canCluster(dict):
+# -*- coding: utf-8 -*-
+import typing
+import canmatrix
+
+
+class CanCluster(dict):
 
     def __init__(self, *arg, **kw):
-        super(canCluster, self).__init__(*arg, **kw)
+        super(CanCluster, self).__init__(*arg, **kw)
+        self._frames = []  # type: typing.MutableSequence[canmatrix.Frame]
+        self._signals = []
+        self._ecus = []
         self.update()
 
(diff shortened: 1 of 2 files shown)


src/canmatrix/cancluster.py
```diff
@@ -1,74 +1,84 @@
...
-    def ecus(self):
+    def ecus(self):  # type: () -> typing.MutableSequence[canmatrix.Ecu]
         if not self._ecus:
-            self.updateECUs()
+            self.update_ecus()
         return self._ecus
 
     @property
-    def boardUnits(self):
+    def board_units(self):  # type: () -> typing.MutableSequence[canmatrix.Ecu]
```
ebroecker: ... while refactoring - pleas rename in `Ecus`

src/canmatrix/formats/kcd.py
```diff
@@ -149,7 +149,7 @@ def dump(dbs, f, **options):
     nodeList = {}
 #    for name in dbs:
 #        db = dbs[name]
-    for bu in canClust.boardUnits:
+    for bu in canClust.board_units:
```
ebroecker: same as above `bu` -> `ecu`

src/canmatrix/cancluster.py
Funth0mas: sure, thanks for hint!

Funth0mas:
The build is failing because of missing `typing` module, it will be resolved by merging #274.

PR #310, opened 2019-03-01
Mypy cleanup

Funth0mas:
- Try to decrease mypy warnings to minimum (CI would need to solve _all of them_)
- Plus some snake_case and BU removal
- Added some "help" to mypy config file
- I silenced (lot of) mypy warnings about "null pointers" - For example the `db.frame_by_name()` might return `None` but is never checked (maybe `raise` an exception instead of returning `None` in the future...)
- There are still about 17 mypy errors, mostly caused by float factory used as `converter`, and 3 real bugs in `compare` module (canmatrix\compare.py:296:8: error: "Frame" has no attribute "extended" ... @ebroecker could you check this please?).
- (And whole `formats` module is not checked yet)

Generally I already see benefits of the type annotations, mypy has actually pointed some bugs.

### DIFF
--- a/src/canmatrix/cancluster.py
+++ b/src/canmatrix/cancluster.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
 import typing
+
 import canmatrix.canmatrix as cm
 
 
--- a/src/canmatrix/cancluster.py
+++ b/src/canmatrix/cancluster.py
@@ -1,20 +1,20 @@
 # -*- coding: utf-8 -*-
 import typing
-import canmatrix
+import canmatrix.canmatrix as canmatrix
 
 
 class CanCluster(dict):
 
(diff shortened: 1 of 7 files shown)


src/canmatrix/cancluster.py
```diff
@@ -1,20 +1,20 @@
 # -*- coding: utf-8 -*-
 import typing
-import canmatrix
+import canmatrix.canmatrix as canmatrix
```
ebroecker: CI seems py27 doesn't like this?

Funth0mas: It seems it's yet another problem :-(  Sorry for spamming with buildless commits, I'll probalby try to download py2 and try it locally.

Funth0mas: Tthe problem was missing `absolute_import` in cancluster.

PR #311, opened 2019-03-02
Move yaml  to format

Funth0mas:
- I hope the `yaml` belongs to `formats`
- Fix start_bit called on signal
- Resolve some Inspection findings

But there are multiple unresolved references in respect to current pyyaml version in pip@py3.


ebroecker:
yes - yaml is a format

PR #318, opened 2019-03-05
Type check with CI

Funth0mas:
Add mypy type check to `tox`, but ignore errorlevel for now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,7 +36,7 @@ commands=
 
 [testenv:mypy]
 description = type check
-basetypthon = python3.6
+basepython = python3.6
 deps=
     mypy
 commands=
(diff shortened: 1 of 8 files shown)


Funth0mas:
retry

Funth0mas:
Okay the mypy runs on travis with py 3.6, see [trace from line 500](https://travis-ci.org/ebroecker/canmatrix/jobs/501993036). The build is (for now) passed regardless of issues found by mypy.

tox.ini
```diff
@@ -33,3 +33,11 @@ commands=
 [testenv:old_tests]
 commands=
     ./test.sh
+
+[testenv:mypy]
+description = type check
+basetypthon = python3.6
```
altendky:
```suggestion
basepython = python3.6
```

Funth0mas:
The minus sign before `python` in tox.ini means "ignore errorlevel".
I found no `allow_failures` for travis (like used for appveyor). Obviously it would be cleaner solution.

Funth0mas:
Hm seems that I'm blind... I'll try to use `allow_failures`

PR #319, opened 2019-03-05
hashed dbc import

timschneider:
created a hashed list of frames to speedup the import process of dbc files, in my example i used a dbc file with 65k lines and around 800 frames.

before ...
 ```
 Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    50600    5.286    0.000    8.465    0.000 canmatrix.py:1390(frame_by_id)
 25100584    3.180    0.000    3.180    0.000 canmatrix.py:600(__eq__)
        1    0.452    0.452    9.610    9.610 canmatrix\formats\dbc.py:437(load)
   110409    0.117    0.000    0.117    0.000 {method 'match' of '_sre.SRE_Pattern' objects}
```

after ...
   ```
Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.388    0.388    1.190    1.190 canmatrix\formats\dbc.py:437(load)
      791    0.131    0.000    0.134    0.000 {built-in method builtins.print}
   110409    0.104    0.000    0.104    0.000 {method 'match' of '_sre.SRE_Pattern' objects}
   759393    0.078    0.000    0.078    0.000 {method 'startswith' of 'str' objects}
```

so it reduced the overall time needed to import that specific dbc file from 9 to 1 seconds on an i7-8550U at 3,9 GHz.

We are able to do this, because the Frame Objects are mutable anyway and stored in the db by reference.

### DIFF
--- a/src/canmatrix/formats/dbc.py
+++ b/src/canmatrix/formats/dbc.py
@@ -450,6 +450,20 @@ def load(f, **options):
     frame = None
     boardUnit = None
     db = canmatrix.CanMatrix()
+    framesById = {}
+    
+    def hash_arbitration_id(arbitration_id):
+        return hash((arbitration_id.id, arbitration_id.extended))
+
+    def get_frame_by_id(arbitration_id):
+        try:
+            return framesById[hash_arbitration_id(arbitration_id)]
+        except KeyError:
+            return None
+            
+    def add_frame_by_id(frame):
+        framesById[hash_arbitration_id(frame.arbitration_id)] = frame
+
     for line in f:
         i = i + 1
         l = line.strip()
@@ -498,6 +512,7 @@ def load(f, **options):
                 frame = canmatrix.Frame(temp.group(2),arbitration_id = int(temp.group(1)),
                                         size=int(temp.group(3)), transmitters=temp.group(4).split())
                 db.frames.append(frame)
+                add_frame_by_id(frame)
             elif decoded.startswith("SG_ "):
                 pattern = r"^SG_ +(\w+) *: *(\d+)\|(\d+)@(\d+)([\+|\-]) +\(([0-9.+\-eE]+),([0-9.+\-eE]+)\) +\[([0-9.+\-eE]+)\|([0-9.+\-eE]+)\] +\"(.*)\" +(.*)"
                 regexp = re.compile(pattern)
@@ -589,7 +604,7 @@ def load(f, **options):
             elif decoded.startswith("BO_TX_BU_ "):
                 regexp = re.compile(r"^BO_TX_BU_ ([0-9]+) *: *(.+);")
                 temp = regexp.match(decoded)
-                frame = db.frame_by_id(canmatrix.ArbitrationId.from_compound_integer(int(temp.group(1))))
+                frame = get_frame_by_id(canmatrix.ArbitrationId.from_compound_integer(int(temp.group(1))))
                 for ecu in temp.group(2).split(','):
                     frame.add_transmitter(ecu)
             elif decoded.startswith("CM_ SG_ "):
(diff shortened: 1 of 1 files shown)


altendky:
@ebroecker, do you think we want to do this as an import-only thing?  Or do you want to explore being able to do this in general?

While it's easy to make things hashable doing so properly often means making the objects immutable.  This is because for dicts etc to work as expected, things that compare equal must hash equal and so the hash must depend on the same things an equality check does.  But the object must also hash consistently otherwise it won't work very well as a dict key.  That's easy enough with attrs classes, just `@attr.s(frozen=True)` and it will be immutable and hashable.  The 'cost' here would be that instead of just modifying a frame's identifier such as `frame.arbitration_id.id = 42` you would `frame.arbitration_id = attr.evolve(frame.arbitration_id, id=42)` (or just make a new one manually, it's a small class `frame.arbitration_id = canmatrix.ArbitrationId(id=42, extended=False)`).

src/canmatrix/formats/dbc.py
```diff
@@ -450,6 +450,20 @@ class FollowUps(object):
...
+
+    def get_frame_by_id(arbitration_id):
+        try:
+            return framesById[hash_arbitration_id(arbitration_id)]
+        except KeyError:
+            return None
+            
+    def add_frame_by_id(frame):
+        framesById[hash_arbitration_id(frame.arbitration_id)] = frame
+
```
altendky: If we don't change the core library to work 'this way' then for a just-import solution I would tend to make a little class (in global scope) I think instead of on-the-fly functions with closures.

```python3
def hash_arbitration_id(...  # as already submitted

@attr.s
class IdFrameMap:
    frames_by_id = attr.ib(factory=dict)

    def get(self, id):
        return self.frames_by_id[hash_arbitration_id(id)]

    def set(self, id, frame):
        self.frames_by_id[hash_arbitration_id(id)] = frame
```

I considered using `__getitem__`/`__setitem__` but since this is really not a proper dict per-se I sort of tend to want it to look a little different in its use.  But, that's a mild tendency at this point, not a hard preference.

ebroecker:
> @ebroecker, do you think we want to do this as an import-only thing? Or do you want to explore being able to do this in general?
> 
> While it's easy to make things hashable doing so properly often means making the objects immutable. This is because for dicts etc to work as expected, things that compare equal must hash equal and so the hash must depend on the same things an equality check does. But the object must also hash consistently otherwise it won't work very well as a dict key. That's easy enough with attrs classes, just `@attr.s(frozen=True)` and it will be immutable and hashable. The 'cost' here would be that instead of just modifying a frame's identifier such as `frame.arbitration_id.id = 42` you would `frame.arbitration_id = attr.evolve(frame.arbitration_id, id=42)` (or just make a new one manually, it's a small class `frame.arbitration_id = canmatrix.ArbitrationId(id=42, extended=False)`).

@altendky 
I also think it would be right to do this general.
The problem @timschneider identified is, that "frame_by_id" is to slow, so this should be fixed.
`@attr.s(frozen=True)` looks good to me at first glance.

timschneider:
@altendky @ebroecker 
I did this as an import only thing because I don't know if anyone is using canmatrix in the way of creating Frames data driven and assigning or changing ids afterwards or if canmatrix wants to support this.

How would you handle this in general if the ArbitrationId class is frozen? Then we have to keep track of the id in the frame or make the id of the frame immutable aswell or do i miss a thing?

altendky:
@timschneider certainly good concerns about how people expect to use this.  I'm not going to suggest we jump to immutable everything...  but @ebroecker is pretty open to change generally.

I think of making the `ArbitrationId` immutable as not a huge step generally.  If it weren't a class and there weren't extended/standard then, it'd be an integer which is immutable anyways.  It seems like it shouldn't be a huge hassle to modify the id of a frame as I showed in my example code above.

But...  I do see your point at the next layer around that.  The dict that presumably we'd create and use for this could use the new hashable `ArbitrationId` as a key but...  the frame could have changed in the meantime and have a new id.  I mean we could make the id a property and so on...  but yeah, creating a system that maintains all relationships accurately is a bit more of a task than shifting your PR here to another part of the code.

I think in general we should consider changes in the direction I proposed but since it's not such a small thing as I originally had in my head, I'll withdraw that proposal from this PR discussion.

---

Any opinions about the class-vs-functions-with-closures comment?
https://github.com/ebroecker/canmatrix/pull/319/files#r263016533

timschneider:
@altendky i see the benefits of a small class and will rewrite this within the next days.

PR #329, opened 2019-03-15
move convert and compare to submodule "cli"

ebroecker:


Funth0mas:
I think that @altendky meant to split the two modules, not only move them. That means leave the "copy" (or "compare" respectively) core functionality in canmatrix/compare.py (for usage as library - from some older PRs we know, that someone uses it as a library already) and only move the `main()` with commandline parser to the sumbodule. (but of course most of modification in the PR are right anyway)

Sorry that I did not remarked, when the PR was open.

ebroecker:
I feared already, but it seemed to me that its the next step to split...
Gonna create another PR

PR #331, opened 2019-03-20
I-Signals with initial value of type float crash conversion and logger levels need improvement

MartyTemple:
Creating a DBC from an ARXML file will crash the canconvert script if an I-Signal has an initial value of type float (e.g. 43000.0 ). The conversion of type text to Int probably needs to include a try and catch exception test however, a simple solution works too.

As for the conversion logger level, I feel the default should show only "INFO" messages, Output Verbosity (-v) should show every status message (debug), and Suppress Status messages (-s) shouldn't show status. The changes that I made in my fork capture the behavior.

Thank you for considering these minor updates.


ebroecker:
Hi @MartyTemple,

Thanks for your PR, gonna have a look on it the next days.

MartyTemple:
Thanks for considering these updates to the Master branch. I am not sure if GitHub made the change to make this merge into your Development branch (or if I neglected to double check the target). However, the main changes can go into both versions.

ebroecker:
merged with #335

PR #336, opened 2019-03-25
move convert and compare to submodule "cli"

ebroecker:
now split em in functionality and library


ebroecker:
@Funth0mas 
is that more the idea of splitting?

Funth0mas:
Yes this is how I would do it too.

ebroecker:
@Funth0mas 
seems ok for merge now, what do you think?

Funth0mas:
I think the code is OK but the old tests seems to need small fix (in fact revert to previous version) to satisfy travis, otherwise we get build failure in development branch I think.

Issue #324, opened 2019-03-07
Add lxml-stubs for xml type checks?


Funth0mas:
The `lxml` library have no annotations by default. But here is a [lxml-stubs](https://github.com/JelleZijlstra/lxml-stubs) project providing them. As the `arxml` Module heavily uses lxml, it might be usefull to include that annotations stubs so that mypy understands the context. 

@altendky  do you think it is a good idea? The project has no pip package, one has to clone it and add to own project.

Funth0mas:
After refactoring the arxml I see that we really need to do it somehow...

altendky:
Sorry I missed this.  :[  I have been pretty booked up and haven't been properly monitoring my email.  It doesn't seem like `canmatrix` should be providing generated stubs for other projects, no.  If there's an existing package it could just be installed as part of a dev package then?  I never did get the `requirements.txt` setup well for `canmatrix` but you can specify `git+https://github.com/JelleZiljlstra/lxml-stubs` in there (or maybe in a new `-dev.txt`?  or maybe this is mostly needed for test?).  Or if they really are generated then just generate them when needed, perhaps.

PR #346, opened 2019-04-04
Annotate and cleanup arxml

Funth0mas:
It is huge renaming hard to review but it was not possible to split to multiple PRs. I tried to rename functions so that they better describe the function objective. If you don't like them, feel free to say it.

I fixed few things when alternatively working with both `lxml.etree.Element` and `ArTree`. Also some small fixes related to renamings in Signal etc. All mypy reported issues in arxml are now indeed issues (`Signal` has no `_initValue` and line 1333 uses wrong data from xml)

Using the old tests I verified that the arxml->csv conversion produces identical result to at least somehow verify that it still works.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -33,5 +33,5 @@ ignore_errors = True
 ignore_errors = True
 
 # other settings:
-[mypy-lxml,xlwt,xlrd,xlsxwriter,past.builtins]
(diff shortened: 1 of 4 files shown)


Funth0mas:
There is million mypy errors in trávit because he don't have lxml stubs #324 :( I'll think how and where to add them into the project.

PR #347, opened 2019-04-07
Add basic lxml stubs

Funth0mas:
Add basic lxml stubs (#324) to allow better arxml type check.
The mypy on travis should only report 5 errors for arxml in this branch.
The stubs folder should't go to distribution package, I hope that ihis way it is excluded from the distribution.


Funth0mas:
it works - mypy: [before](https://travis-ci.org/ebroecker/canmatrix/jobs/516359867), [after](https://travis-ci.org/ebroecker/canmatrix/jobs/516957793)

Issue #367, opened 2019-05-31
Issue in converting an excel file to dbc


Jayanthdj:
Could you please help as I am facing issue in converting an excel files to dbc. I am attaching the screenshot
First file error
![image](https://user-images.githubusercontent.com/33441532/58683436-05f64400-8392-11e9-9f49-8296cc000cdb.png)
 second file error
![image](https://user-images.githubusercontent.com/33441532/58683464-24f4d600-8392-11e9-9e8c-339bc51a467c.png)

Funth0mas:
Could you please provide minimal example of such excel file?

yuren1986:
Hello Funth0mas,

 I also find the same issue report.

![image](https://user-images.githubusercontent.com/51859118/59553539-caca5680-8fc8-11e9-8d50-d588492eec5c.png)

So can we use the canconvert change the .xlsx file to .dbc ?

Hope your feedback, thanks.

Funth0mas:
Hi @yuren1986, both issues are pretty unrelated. The first has "incompatible" header whereas your table contains spaces in the cell, where canmatrix expects CAN ID ("ID" in the header). It must either contain valid hexadecimal CAN ID ended by "xh" or empty cell (really empty, not containing spaces not visible by eyes)

You can find example table under https://github.com/Funth0mas/canmatrix/blob/development/test/test.xls and https://github.com/Funth0mas/canmatrix/blob/development/test/test.xlsx  I see possible discrepancy with plain non-extended can IDs

If it doesn't help you, let provide your table (simplified, with modified data if they're secret) for analysis.

yuren1986:
Hello Funth0mas,
Thanks for your helping. I had checked your suggestion the test.xlsx can be transformed to dbc file, but the dbc file can't open.

The error as following, i think the reason is caused by the send type. 
![image](https://user-images.githubusercontent.com/51859118/59564476-f2c5c280-9079-11e9-880e-6a01857d7eb3.png)

And I use the TEST file cmTemplate.xlsx change to the .dbc , and I found the CAN ID has been changed from 123h to 12h.
![image](https://user-images.githubusercontent.com/51859118/59564634-b004ea00-907b-11e9-98f2-4dd01705439b.png)

So maybe the data information has been changed wrong by data type between transformation.]

Do you find the same issue also?

ebroecker:
Looks like the define ist Missing. I'll have a Look on it

Funth0mas:
[...]

PR #377, opened 2019-07-12
fix: xls is using the wrong number for arbitration id

coccyx00:
When loading xls, the arbitration id is misinterpeted by one character. This leads to wrong ids.


ebroecker:
thanks @coccyx00  for your PR.

this is a duplicate of issue #367 (with branch iss367 for the fix)

as you did the same fix like me (in branch iss367) I see this as verification and gonna merge.

Issue #379, opened 2019-07-30
Wrong type in json


chrisoro:
Here https://github.com/ebroecker/canmatrix/issues/102 not only the internal representation but also the json export itself were changing.
That lead to 
```json
"factor": str(signal.factor),
"offset": str(signal.offset),
"min": str(signal.min),
"max": str(signal.max),
```

In my opinion it is okay to handle it internally as a different type but json allows floating point numbers, so why not use it? The change basically requires everyone using such an exported json to convert the string back to float.

May I ask why this was changed? Any chance to revert the export back to using float instead of string?

altendky:
The issue is that floats are inaccurate so I use `decimal.Decimal`.  They are unable to correctly represent values like 1.1.  Though I don't use the JSON output so this exact code isn't directly because of me, but still maybe indirectly.  But sure, floats are so prevalent that at least my first thought is that it would be reasonable to have that controllable in the JSON output.  Should be easy enough to add an option to `canmatrix.formats.cmjson.dump()`.  Or maybe separate options per output?  Maybe, probably not?  Anyways, an option where you would pass a callable such as `str` or `float`.  And yes, I'm sure the `float_factory` thing is still a pain for everyone here but since this is an output variation instead of a core data type it shouldn't leak all over the place in uncontrollable ways, hopefully.

Sorry, but I probably won't manage to get to this soon myself.  Hopefully one of the other contributors (or you `:]`) have time for  PR.

chrisoro:
Sure, I can take care of it. I just wanted to make sure what exactly you guys have in mind since I don't understand ever needing to export a string instead of float/int for a signal.
offset, min and max should be fine as int
factor should be fine as float

I would just replace that str() with the appropriate type, at least for json dump(). In load() you can still convert it to a string for internal handling.  You ok with that?

ebroecker:
@chrisoro 
this would be nice.
Otherwiese I'll find some time to fix this.

altendky:
I'm not clear why integers are sufficient for any of them vs float.

altendky:
I'm not clear why integers are sufficient for any of them vs float.

altendky:
People expect floats.  People also don't understand floats.  It's not obvious to me that choosing an output format (float rather than string) that is unable to actually represent the internal values (of type `decimal.Decimal` by default) is a good thing.

https://repl.it/@altendky/FrightenedWrongCompilerbug-1
```python3
import decimal

def roundtrip(original, serializer):
[...]

PR #380, opened 2019-07-31
Dump factor, offset, min, max as float. Load as numeric

chrisoro:
https://github.com/ebroecker/canmatrix/issues/379

### DIFF
--- a/src/canmatrix/formats/cmjson.py
+++ b/src/canmatrix/formats/cmjson.py
@@ -42,6 +42,7 @@ def dump(db, f, **options):
     export_canard = options.get('jsonCanard', False)
     motorola_bit_format = options.get('jsonMotorolaBitFormat', "lsb")
     export_all = options.get('jsonAll', False)
+    native_types = options.get('jsonNativeTypes', False)
     additional_frame_columns = [x for x in options.get("additionalFrameAttributes", "").split(",") if x]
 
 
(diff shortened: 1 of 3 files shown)


src/canmatrix/formats/cmjson.py
```diff
@@ -42,6 +42,7 @@ def dump(db, f, **options):
     export_canard = options.get('jsonCanard', False)
     motorola_bit_format = options.get('jsonMotorolaBitFormat', "lsb")
     export_all = options.get('jsonAll', False)
+    native_types = options.get('jsonNativeTypes', False)
```
altendky:
```python
number_converter = float if native_types else str
```
Then below we can just unconditionally
```python
"factor": number_converter(signal.factor)
```
and similar.

src/canmatrix/formats/cmjson.py
```diff
@@ -193,11 +194,12 @@ def load(f, **_options):
...
                     is_little_endian=is_little_endian,
                     is_signed=is_signed,
-                    is_float=is_float,
-                    factor=signal["factor"],
-                    offset=signal["offset"]
+                    is_float=is_float
                 )
 
+                new_signal.factor = new_signal.float_factory(signal["factor"])
+                new_signal.offset = new_signal.float_factory(signal["offset"])
```
altendky: Was `Signal` not properly applying the `float_factory` itself?

chrisoro: Oh, I didn't catch that. Thanks

src/canmatrix/formats/cmjson.py
chrisoro: Sure.

Issue #368, opened 2019-06-07
Value type 'is_signed': converting arxml to dbc


Mombox:

ilazrak:
@ebroecker 

'unsigned signals' in arxml become 'signed signals' after conversion to dbc,
which also affects the Min and Max values of the obtained signals...

        rawRange = 2 ** (self.size - (1 if self.**is_signed** else 0))
        return (
            factory(-rawRange if self.**is_signed** else 0),
            factory(rawRange - 1),

Funth0mas:
Could you please provide minimal example of such arxml file?

ilazrak:
[NewCluster_max_min.zip](https://github.com/ebroecker/canmatrix/files/3319776/NewCluster_max_min.zip)

@Funth0mas 
attached one Frame similar to my arxml data

thanks

Funth0mas:
Okay I accidentally reproduced the same issue with own ARXML :-) @ebroecker, there is some signed/unsigned guessing based on `BASE-TYPE` name. Instead, as described in https://www.autosar.org/fileadmin/user_upload/standards/adaptive/17-10/AUTOSAR_SWS_AdaptivePlatformTypes.pdf, we shall check for `<BASE-TYPE-ENCODING>` value:

 * `NONE` for unsigned integers
 * `2C` for signed integers "two's complement"
 * `IEEE754` for float32/float32
 * `BOOLEAN`

have you already some progress, or shall I try to fix it? I might find some time to do it.

ebroecker:
@Funth0mas 
You are probably right...
Well - you are welcome - if you find time - do fix this

Funth0mas:
@ilazrak please does the linear scale used in you example file make sense? There are
```
                      <COMPU-NUMERATOR>
                        <V>0</V>
                        <V>1</V>
                      </COMPU-NUMERATOR>
                      <COMPU-DENOMINATOR>
                        <V>0</V>
                        <V>1</V>
```
If I understand correctly that is a polynomial (0 + 1*x)/(0 + 1*x) = 1, that means constant one... unfortunately this also leads to divide-by-zero in canmatrix because it expects such function to be expressed simply as (1 + 0*x)/(1 + 0*x) => offset=1/1=1, scale=0/1=0.

ebroecker:
@iartez could you please try branch iss385?

Thanks

ebroecker:
fixed in development now.
Please reopen or open new issue if still not working as expected

[...]

PR #390, opened 2019-08-14
Iss385 // Arxml Container support and ARXML refactoring

ebroecker:


src/canmatrix/formats/arxml.py
```diff
@@ -927,9 +927,12 @@ def decode_compu_method(compu_method, root_or_cache, ns, float_factory):
             numerator = get_children(numerator_parent, "V", root_or_cache, ns)
             denominator_parent = get_child(rational, "COMPU-DENOMINATOR", root_or_cache, ns)
             denominator = get_children(denominator_parent, "V", root_or_cache, ns)
-
-            factor = float_factory(numerator[1].text) / float_factory(denominator[0].text)
-            offset = float_factory(numerator[0].text) / float_factory(denominator[0].text)
+            try:
+                factor = float_factory(numerator[1].text) / float_factory(denominator[0].text)
+                offset = float_factory(numerator[0].text) / float_factory(denominator[0].text)
+            except decimal.DivisionByZero:
```
Funth0mas: This is due to example provided in #368, right? I believe that the "V"alues in the (DE)NOMINATOR are wrong, because they should represent coefficients of a polynomial function and it would be (0+x)/(0+x) = 1 regardless of the signal value (and undefined for for x=0). Strange.
I'd write at least an error message to the console, that the polynom is not supported and it is replaced by factor=1.

ebroecker: @Funth0mas 
Thanks for this hint.
Fixed this with f29af5a

PR #406, opened 2019-10-17
update arxml.py

yangzh1:
delete "ns+".
With "ns +", arxml.py cann't parse arxml version 4.2.0 rignt.

### DIFF
--- a/src/canmatrix/formats/arxml.py
+++ b/src/canmatrix/formats/arxml.py
@@ -1397,7 +1397,7 @@ def get_frame(frame_triggering, root_or_cache, multiplex_translation, ns, float_
 
     store_frame_timings(new_frame, cyclic_timing, event_timing, minimum_delay, repeats, starting_time, time_offset, repeating_time, root_or_cache, time_period, ns, float_factory)
 
-    if pdu.tag == ns + "CONTAINER-I-PDU":
+    if pdu.tag == "CONTAINER-I-PDU":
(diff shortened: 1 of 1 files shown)


ebroecker:
Are you sure, this is an issue here and not in your ARXML-File? 
Do you have a complete example?

It will break all older ARXMLs, because namespace was is always needed.

yangzh1:
> Are you sure, this is an issue here and not in your ARXML-File?
> Do you have a complete example?
> It will break all older ARXMLs, because namespace was is always needed.

It's my first time to parse arxml version 4.2.0,maybe I'm wrong.But with canconvert,I can't parse arxml  version4.2.0 right. 
In my arxml version 4.2.0 file, the arxml PDU statement is:
<PDU-REF DEST="CONTAINER-I-PDU">/Communication/Pdu/HCU_HSC6_M10_FrP00</PDU-REF>
So,if  the parser is "if pdu.tag == ns + "CONTAINER-I-PDU":", can we goto the if-statement? What does the "ns" mean?

ebroecker:
Hi,
the `ns` means *Name Space* it's usually in the top of the ARXML

something like
```XML
<AUTOSAR xsi:schemaLocation="http://autosar.org/schema/r4.0 AUTOSAR_4-2-2.xsd" xmlns="http://autosar.org/schema/r4.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```

ebroecker:
@yangzh1 

Did you find some solution? Or do we have some strange behavior in canmatrix?

PR #445, opened 2019-12-18
Fix some bugs in FIBEX export

akaanich-technica:
## Changes:
* Fixed duplicate ids when two signals have same name in different frames
* Added CONNECTORS element
* Fixed location of PROCESSING-INFORMATION
* Added the bit length of each signal

### DIFF
--- a/src/canmatrix/formats/fibex.py
+++ b/src/canmatrix/formats/fibex.py
@@ -172,9 +172,10 @@ def dump(db, f, **options):
         create_sub_element_fx(pdu, "PDU-TYPE", "APPLICATION")
         signal_instances = create_sub_element_fx(pdu, "SIGNAL-INSTANCES")
         for signal in frame.signals:
+            signal_id = frame.name + "." + signal.name
             signal_instance = create_sub_element_fx(
                 signal_instances, "SIGNAL-INSTANCE")
-            signal_instance.set("ID", "PDUINST_" + signal.name)
+            signal_instance.set("ID", "PDUINST_" + signal_id)
             # startBit: TODO - find out correct BYTEORDER ...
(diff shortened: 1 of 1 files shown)


akaanich-technica:
Failure to due: `The job exceeded the maximum log length, and has been terminated.`

ebroecker:
@ akaanich-technica 

thanks for this PR.
So FIBEX code is so usable for you? 
I once implemented it only from white paper - without any chance to test...

akaanich-technica:
@ebroecker It is usable:
We used the DBC files from https://github.com/commaai/opendbc and converted them to FIBEX, and consume in our own toolchain.

The only problem that I couldn't 100% fix is that some DBC files had duplicate signal names, duplicate frame names, and sometimes even duplicate ecu names.

The one thing I am sure of, is that it is valid in DBC to have two different signals with the same name, as long as they reside in different frames, so I fixed that in the FIBEX exporter, the rest I wasn't so sure about.

ebroecker:
Hi @akaanich-technica ,

thanks for pointing to opendbc - didn't know this...

Most tools won't allow signals with equal names. You could even use canmatrix to fix this:
```
python -m canmatrix.cli.convert source.dbc target.dbc
```
should create a dbc with uniq signal names (`--no-dbcUniqueSignalNames` is the option to disable this behaviour)

In second step convert to fibex as before...

akaanich-technica:
@ ebroecker Thanks for the tip,
I see this PR is failing App Veyor, but I can't pinpoint what test cases are failing.
Is this PR failing due to my changes, or something off with App Veyor configuration?

akaanich-technica:
@ebroecker AppVoyer is failing in the development branch as well, so those changes should be good to merge, no?

PR #466, opened 2020-03-26
Adding changes to allow ArbitrationId classes to be sorted

sou1hacker:
Changed default arbitration_id.extended to False from None
Removed the existing equals method and using attr.s default comparison methods
The current logic implemented is that:
 ArbitrationId(id=0x1, extended=False) <   ArbitrationId(id=0x1, extended=True)
 ArbitrationId(id=0x1, extended=True) < ArbitrationId(id=0x2, extended=False)

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -35,6 +35,7 @@ import logging
 import math
 import struct
 import typing
+import warnings
 from builtins import *
 
 import attr
@@ -580,7 +581,7 @@ def pack_bitstring(length, is_float, value, signed):
     return bitstring
 
 
-@attr.s(cmp=False)
+@attr.s
 class ArbitrationId(object):
     standard_id_mask = ((1 << 11) - 1)
     extended_id_mask = ((1 << 29) - 1)
@@ -593,7 +594,11 @@ class ArbitrationId(object):
         if self.extended is None:
             # Mimicking old behaviour for now -- remove in the future
             self.extended = True
-            logger.warning("Please set 'extended' attribute as a boolean")
+            warnings.warn(
+                "Please set 'extended' attribute as a boolean instead of "
+                "None when creating an instance of ArbitrationId class",
+                DeprecationWarning
+            )
         if self.extended:
             mask = self.extended_id_mask
         else:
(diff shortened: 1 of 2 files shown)


pierreluctg:
@altendky and @ebroecker, please have a look.

Thank you

src/canmatrix/canmatrix.py
```diff
@@ -712,14 +716,7 @@ def to_compound_integer(self):
             return self.id
 
     def __eq__(self, other):  # type: (ArbitrationId) -> ArbitrationId
-        return (
```
ebroecker: the original intention of _None_ was to create the possibility to compare an arbitration id while ignoring extended flag.
I think, there are use cases where people want to find a arbitration id 5 and they don't bother if its extended or regular

src/canmatrix/canmatrix.py
```diff
@@ -587,10 +587,14 @@ class ArbitrationId(object):
...
     id = attr.ib(default=None)
-    extended = attr.ib(default=None)
+    extended = attr.ib(default=False)  # type: bool
 
     def __attrs_post_init__(self):
-        if self.extended is None or self.extended:
+        if self.extended is None:
+            # Mimicking old behaviour for now -- remove in the future
+            self.extended = True
+            logger.warning("Please set 'extended' attribute as a boolean")
```
altendky: This should be a warning rather than being logged.  I'm not sure but maybe there's something in https://pypi.org/project/regret/ that would take it to a higher level yet.

src/canmatrix/canmatrix.py
altendky: What about `a.id == b.id` for that?

src/canmatrix/canmatrix.py
```diff
@@ -711,15 +715,36 @@ def to_compound_integer(self):
...
-        )
+    def __eq__(self, other):  # type: (ArbitrationId) -> ArbitrationId
+        return self.id == other.id and self.extended == other.extended
+
+    def __le__(self, other):
+        if not self.extended and other.extended:
+            return True
+        if self.extended and not other.extended:
+            return False
+        return self.id <= other.id
```
```suggestion
        return (self.extended, self.id) <= (other.extended, other.id)
```

But can this not be done via some attrs configuration and we could remove `cmp=False`?

src/canmatrix/canmatrix.py
sou1hacker: I do feel like we need to remove None as a possible option for extended. Otherwise the following will be True which may cause issues during sorting, etc.:
Abr(extended=True) == Abr(extended=None) == Abr(extended=False) but Abr(extended=True) != Abr(extended=False)

src/canmatrix/canmatrix.py
sou1hacker: Done

src/canmatrix/canmatrix.py
sou1hacker: Done

sou1hacker:
@altendky @ebroecker Reverted comparison functions so that default ones are used. This should make them more standardized and the behavior more predictable. Please re-review.

pierreluctg:
Hi @ebroecker and @altendky, any concern merging this PR?

PR #468, opened 2020-04-07
Check for EOF when parsing enums in a .sym file. Fixes #465

### LINKED ISSUE #465: Unterminated enum line in .sym results in infinite loop
```
FormatVersion=5.0 // Do not edit this line!
Title="AFE_CAN_ID0"

{ENUMS}
enum Relay(0="Open, 1="Closed",2="Error",3="N/A"
```

https://github.com/ebroecker/canmatrix/blob/180829012a8aee7e176f4f5b4c039c7e7989e8f5/src/canmatrix/formats/sym.py#L354-L359

https://docs.python.org/3.8/library/io.html?highlight=readline#io.TextIOBase.readline
> "If the stream is already at EOF, an empty string is returned."

So we need to check for an empty string from `.readline()` and raise some exception.

Seneda:
I have added a fix for #465 and a new test case in test_sym.py that checks for an EOFError in the case of an untemrinated enum.

### DIFF
--- a/src/canmatrix/formats/sym.py
+++ b/src/canmatrix/formats/sym.py
@@ -354,9 +354,12 @@ def load(f, **options):  # type: (typing.IO, **typing.Any) -> canmatrix.CanMatri
                     while not line[5:].strip().endswith(')'):
                         line = line.split('//')[0]
                         if sys.version_info > (3, 0):  # is there a clean way to to it?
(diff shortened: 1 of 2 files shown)


Seneda:
Looks like the exeption works slightly differently in python2.7, I will correct this

PR #479, opened 2020-04-28
Parse sym types

### LINKED ISSUE #477: Cannot parse "Char", "String" or "Raw" signal types in .Sym files
The PCAN Symbol editor allows several types to be set for signals. These are:

- Bit 
- Char
- String
- Signed
- Unsigned
- Float
- Enum
- Double 
- Raw

Currently Canmatrix will accept some of these but not all. 
The ones which do not work yet are:

- Char
- String
- Enum
- Raw

I think that the first thing to do is to make it interpret all of these except Enum as unsigned. So that it doesn't throw any signals away.
I think it would also be useful to add a property to indicate what type the symbol file specified the signal to be, so that a symbol file can be written back with the same types as it was read with.

(Adding Enums is a bit more complex, so ill open a separate issue for those.)

Seneda:
Fixes #477 

Allow parsing of the main .sym signal types : string, raw, char.

Add a type_label attribute so that symbol files can be saved back to .sym format without loss of information.
Add an is_ascii attribute for symbols Add a check to see if Signal names need Quoting when saving to a .sym file (if they start with a number or contain a space etc.)

Add a test which reads in a .sym file containing all signal types and checks that they are all correctly imported and exported.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -155,6 +155,8 @@ class Signal(object):
 
     mux_value = attr.ib(default=None)
     is_float = attr.ib(default=False)  # type: bool
+    is_ascii = attr.ib(default=False)  # type: bool
+    type_label = attr.ib(default="")
     enumeration = attr.ib(default=None)  # type: typing.Optional[str]
     comments = attr.ib(factory=dict)  # type: typing.MutableMapping[int, str]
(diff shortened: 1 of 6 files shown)


Seneda:
@ebroecker Do you see any problems with how I've implemented this? I'm happy to make any changes if you don't think this should be merged as it is.

PR #527, opened 2020-11-12
src/canmatrix.py: fix nested mux with no values

rfrowe:
If a complexly multiplexed signal has a sub-multiplexer with no values, a TypeError is thrown.

I ran into this while using [commaai/opendbc](https://github.com/commaai/opendbc) which muxes first on OBD-II service and then PID. They define a service 2 but do not enumerate any PIDs for it, causing the above referenced exception.


ebroecker:
@rfrowe  Thanks a lot!

PR #532, opened 2020-12-03
add is_fd flag for json output

DennisRoeck:
fd flag is implemented but not used in json.
add is_fd flag to the json exporter.


ebroecker:
Thanks @DennisRoeck

PR #552, opened 2021-03-20
Correctly the result if compare two frame with different ID

anphucnguyen:
When comparing two dbc, the the frame has different ID it should be marked as "changed" frame.


ebroecker:
Thanks!

PR #564, opened 2021-04-10
arxml fix

leixao:
Fix some problems in arxml conversion


src/canmatrix/formats/arxml.py
```diff
@@ -1231,7 +1231,7 @@ def get_signals(signal_array, frame, ea, multiplex_id, float_factory, bit_offset
 
             if initvalue is not None and initvalue.text is not None:
                 initvalue.text = canmatrix.utils.guess_value(initvalue.text)
-                new_signal.initial_value = float_factory(initvalue.text)
```
ebroecker, line 1234: I am in discussion, if raw_value or physical value should be stored in intial_value (#567) 
so we should wait here

PR #566, opened 2021-04-12
fix initial value setting

jazi007:
use min value when initial_value default value is out of range [min, max]


ebroecker:
ok, now we need to fix the unit-tests...

jazi007:
Unit test fixed

PR #567, opened 2021-04-14
Set initial_value to physical value

pierreluctg:
Currently the initial_value is incorrectly set to the signal raw value.

### DIFF
--- a/src/canmatrix/formats/dbc.py
+++ b/src/canmatrix/formats/dbc.py
@@ -233,8 +233,8 @@ def dump(in_db, f, **options):
         for signal in frame.signals:
             if signal.cycle_time != 0:
                 signal.add_attribute("GenSigCycleTime", signal.cycle_time)
-            if signal.initial_value != 0:
-                signal.add_attribute("GenSigStartValue", signal.initial_value)
+            if signal.phys2raw(None) != 0:
+                signal.add_attribute("GenSigStartValue", signal.phys2raw(None))
 
             name = normalized_names[signal]
             if compatibility:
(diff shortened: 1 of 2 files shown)


ebroecker:
Hi @pierreluctg,

why is raw value as initial_value wrong in your eyes?
So you think initial_value should be physical value.
I thought it should be the raw value, because with this we don't have problems regarding scaling.

pierreluctg:
Hi @ebroecker, based on the comment/type hint here https://github.com/ebroecker/canmatrix/blob/development/src/canmatrix/canmatrix.py#L173, `initial_value` is a `canmatrix.types.PhysicalValue`

It's also a physical value according to the logic in `phys2raw` here https://github.com/ebroecker/canmatrix/blob/development/src/canmatrix/canmatrix.py#L407 else the `intitial_value` would be returned directly without scaling.

We can change the type hinting and logic to have `initial_value` as a `canmatrix.types.RawValue` if that is your preference.

ebroecker:
@pierreluctg 
I feared already that it's inconsistent. 
I don't want to prescribe something. 
Just want to discuss.
My only argument is scaling.  But as far as I remembre, there are formats which use physical values and formats wich use raw value.

So anyway, what du You think ist the more intuitive way to to it?

pierreluctg:
IMO, the `initial_value` attribute should be a physical value, just like `min`, `max`, etc.
From a canmatrix API point of view it also make sense to have all the attributes as physical values.

This is in line with how application like Vector CANdb++ is displaying the information.

This also make the distinction between the DCB attribute `GenSigStartValue` (raw initial value) and `initial_value` (physical initial value).

ebroecker:
Hi @pierreluctg 
thanks for clarification!

PR #569, opened 2021-05-03
arxml CAN parsing optim and fixes

jazi007:
### optimize parsing
170MB file parsing goes from 13 minutes to 3 minutes
### ECU extraction
fix ECU receivers/transmitters per cluster

### DIFF
--- a/src/canmatrix/formats/arxml.py
+++ b/src/canmatrix/formats/arxml.py
@@ -27,6 +27,7 @@
 from __future__ import absolute_import, division, print_function
 
 import decimal
+import copy
 import logging
 import typing
 from builtins import *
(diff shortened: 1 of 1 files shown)


jazi007:
Hello @ebroecker,

some changes done in the arxml parser.

I will try to create some issue for other comments I have.

those changes are mainly related to speed up parsing for large ARXML files and to match the correct ECU (sender/receivers) when export do dbc is done. (as in dbc you are getting first ecu in the list for transmistter)

let me know if it's ok for you

thanks

ebroecker:
Hi @jazi007 

for sure it is ok for me!

thanks a lot.

PR #586, opened 2021-07-23
Fix typo in cli/convert.py file

AjinkyaPasalkar:
It looks like line 88 was copied from line 87, but while editing '--no-arxmlFlexray' was not updated.
I have simply changed --no-arxmlFlexray to --no-arxmlEthernet
@ebroecker


ebroecker:
thanks

PR #590, opened 2021-07-28
bugfix: no signals available for a pdu

tainnok:
the list of signals (inside the PDUs) are always empty.
to get the signals for a pdu you have to use follow_all_ref instead of follow_ref


tainnok:
mixed 2 pull-requests, closed this one and created two seperate ones

PR #593, opened 2021-07-28
arxml: read data related to E2E-protection

tainnok:
in arxml there are informations stored how the End2End-Protection need to be done, which is needed to transmitt and receive E2E-protected Signalgroups.

so i added a basic implementation to read some basic informations, until now the E2E-Profile and the data-ids which are needed during calculation of the E2E-protection.

this informations stored at the I-SignalGroup, but until now even if i I-Signalgroup was found, the System-Signal-Group was searched and used.
I changed it in the way that if the I-Singalgroup was found also this Signalgroup will be used (and renamed the method according).

now, if a I-Signalgroup is added which contains END-TO-END-TRANSFORMATION-I-SIGNAL-PROPS it will follow the TRANSFORMER-REF to get the correct E2E-Profile and store this together with the data-ids into a dictionary which will be added to the Signalgroup itself.

for me thats the most simple way to implement this.

BUT: imho the fact that there are different kinds of SignalGroups available in the arxml should also be refelcted by canmatrix, so imho canmatrix should read the System-Signalgroup AS WELL as the I-Signalgroup and store both into the canmatrix object.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -49,7 +49,8 @@ import canmatrix.utils
 if attr.__version__ < '17.4.0':  # type: ignore
     raise RuntimeError("need attrs >= 17.4.0")
 logger = logging.getLogger(__name__)
-defaultFloatFactory = decimal.Decimal  # type: typing.Callable[[typing.Any], canmatrix.types.PhysicalValue]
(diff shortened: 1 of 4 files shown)


tainnok:
closed because pull-request contains other/not related changes too

PR #624, opened 2022-02-11
[dbc] allow multiple space after : in Frame definition

ebroecker:


ALUE:

PR #637, opened 2022-04-04
Added Switches checkFrameNameLength, checkSignalNameLength

Arcola-Kankan:
1. CANdb editor does not allow frame or signal names  more than 32 character long , resulting in error message when frame is opened in the editor. Hence _**checkFrameNameLength**_, _**checkSignalNameLength**_ switches are implemented to warn against frames or signals with names more than 32 characters. 
2. Added _**recalcSignalMinimums**_ switch to force recalculation of signal minimum when set to enabled.
3. Renamed below switches 
- **_calcSignalMax_** renamed to **_recalcSignalMax_**
- **_recalcSignalMax_** renamed to **_recalcSignalMaximums_**
- **_checkSignalRange_** renamed to **_warnSignalMinMaxSame_**


ebroecker:
Hi @Arcola-Kankan,

how does the `frame name length > 32` chars work together with the meachnism around `SystemMessageLongSymbol` (https://github.com/ebroecker/canmatrix/blob/development/src/canmatrix/formats/dbc.py#L204) where we use the dbc way to handle long frame names?

Till now, I thought that `SystemMessageLongSymbol` handles the long frame names... Maybe I do understand something wrong here?

Arcola-Kankan:
Hi @ebroecker , the purpose of the switches are to warn the user when running the can convert script that the frame name or the signal name is exceeding 32 char limit. So that the user can make changes accordingly.

Though the output dbc has the attribute SystemMessageLongSymbol/SystemSignalLongSymbol , the candb editor still complains about the frame/signal name being long
image

ebroecker:
Hi @Arcola-Kankan ,

Thanks for the quick reply, I'll have to check this on my side.

Arcola-Kankan:
Thanks @ebroecker

ebroecker:
you can change this in candb++ and increase the allowed length to 128

grafik

ebroecker:
@Arcola-Kankan,

Ok, if you want just warnings, than move them to the length-check as you wrote above (https://github.com/ebroecker/canmatrix/blob/development/src/canmatrix/formats/dbc.py#L204). This makes sense!

Maybe it would be nice to have an option which allows to shorten to long names also, but this is just an idea.

Arcola-Kankan:
Hi @ebroecker , I will add the warnings in the mentioned place. 

> Maybe it would be nice to have an option which allows to shorten to long names also, but this is just an idea.
I did not get this , can you please explain ?

Arcola-Kankan:
Hi @ebroecker . There was a missing newline conflict on my branch , which i have updated . Please proceed with the merge.

Regards,
kankan

PR #641, opened 2022-05-17
bugfix float_factory: there is no base-argument

tainnok:
bug since commit 97ebe53c66556410d9484f7335c6702f24c19b5f
the float_factory - doesn't matter if Deciaml() or float() is used - does not have a base-argument like int()


ebroecker:
Hi @tainnok 
thanks a lot for fixing my mistage

PR #642, opened 2022-05-17
bugfix: get all receivers/transmitters and not only one

tainnok:
bug since commit d2fbdaae8ef622deff26054bfa497c55ff30779d

before commit d2fbdaae8ef622deff26054bfa497c55ff30779d for each port in port_frames the ecu was checked and added as receiver or transmitter.
after (bug) only the first comm_direction was added as receiver or transmitter, the remaining ones are missed in the list of receivers/transmitters.


ebroecker:
@tainnok 

thanks again for fixing my bugs!

PR #647, opened 2022-07-23
arxml - pdu name

### LINKED ISSUE #646: no change showed when compare two arxml( i changed the PDU name of arxml)
when i changed the PDU name in the arxml and tried to compare the difference.   the result didn't show the PDU name change 
 i use cmd like this:cancompare test_bak.arxml test.arxml
i recommand to show the PDU name change  cause when import the arxml to AUTOSAR Tool Chain  the PDU name matters.
图片]

ebroecker:
add pdu-name and can-frame-triggering-name to frame attributes during arxml read
add option for convert to rename frame form frame.attribute

should fix #646 also


ALUE:

PR #658, opened 2022-10-13
add pdu_name to frame which represents the pdu-name of the pdu-frame-…

tainnok:
Until now the Name of the PDU which is mapped to the frame is not accesible via CanMatrix.
Now add a attribute 'pdu_name' to the Class Frame which represents the name of the PDU which is mapped to the Frame.

Hint:
In arxml it's possible that more than one PDU is mapped to a frame, but this is not supported at all in CanMatrix, so this pdu-name is also not a list.
This does NOT conflict with Container-PDU's (where the name is already available).


ebroecker:
Thanks!

PR #659, opened 2022-10-13
precise logging related to Frame/Frametrigger to avoid confusions

tainnok:
just a few changes regarding logging to prevent confusing situations (Frame <-> Frame-Trigger)


ebroecker:
thanks

PR #662, opened 2022-10-17
support missing information like channel_id, included-pdu_id and conn…

rawiamoalla:
…ector_id

### DIFF
--- a/src/canmatrix/formats/fibex.py
+++ b/src/canmatrix/formats/fibex.py
@@ -101,7 +101,7 @@ def dump(db, f, **options):
     cluster.set('ID', 'canCluster1')
     create_short_name_desc(cluster, "clusterShort", "clusterDesc")
     create_sub_element_fx(cluster, "SPEED", "500")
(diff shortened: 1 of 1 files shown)


rawiamoalla:
Could you please merge this merge request

src/canmatrix/formats/fibex.py
```diff
@@ -101,7 +101,7 @@ def dump(db, f, **options):
     cluster.set('ID', 'canCluster1')
     create_short_name_desc(cluster, "clusterShort", "clusterDesc")
     create_sub_element_fx(cluster, "SPEED", "500")
-    create_sub_element_fx(cluster, "IS-HIGH-LOW-BIT-ORDER", "false")
```
ebroecker, line 104: hi  @rawiamoalla 

are you sure here?  I did not double check, so just a question.

This is a change - the rest of your patch is no problem.

rawiamoalla: Hi @ebroecker normally yes. there is no problem if we replaced to true

ebroecker: ok, perfect

PR #691, opened 2023-02-28
fix: load json attributes from each message

russkel:
Hi Ed,

I noticed the message attributes were not being loaded for jsons. Here is a fix for that.

Russ


ebroecker:
Thanks

Issue #699, opened 2023-04-20
DeprecationWarning for accessing `attr.__version__`


timothygebhard:
[Line 49](https://github.com/ebroecker/canmatrix/blob/b5abc6982ee1b1e52b395b5d93039df73c379ac2/src/canmatrix/canmatrix.py#L49) in `canmatrix.py` is giving me a DeprecationWarning:

```
DeprecationWarning: Accessing attr.__version__ is deprecated and will be removed in a future release. Use importlib.metadata directly to query for attrs's packaging metadata.
    if attr.__version__ < '17.4.0':  # type: ignore
```

I think fixing it should be as easy as:

```python
from importlib.metadata import version

if version("attr") < '17.4.0':  # type: ignore
    raise RuntimeError("need attrs >= 17.4.0")
```

_However_, the [`attr` package on PyPI](https://pypi.org/project/attr/) currently only is at version 0.3.2, so this will _always_ give a RuntimeError. Additionally, the error message refers to `attrs` (note the `s` at the end), and that's also what is required [in the setup.py](https://github.com/ebroecker/canmatrix/blob/b5abc6982ee1b1e52b395b5d93039df73c379ac2/setup.py#L83), so maybe it should be `attrs` everywhere instead of `attr`? 🙂

ebroecker:
@timothygebhard 
thanks for your report.

ebroecker:
should be fixed

PR #700, opened 2023-04-25
fix issue #699

ebroecker:
fix attr version check issue


ALUE:

werux:
Hi @ebroecker, I just noticed that this would only work on Python 3.8+, right?

PR #727, opened 2023-10-13
Ignore SecuredIPdus with no Payload

werux:
Log SecuredIPdus without Payload and keep processing frames.


werux:
Hi @ebroecker, can you please help me review this one?

Issue #731, opened 2023-10-31
Signal 's default value is not set when converting from DBC to FIBEX


welbouri:
When converting a DBC file containing some signals' default values to a FIBEX file, `DEFAULT-VALUE`  is missing for signals.

This can be noticed when checking with this [file](https://github.com/ebroecker/canmatrix/blob/c956871250ba06f753dd7477d3dde37932a89ed7/test/reference/from_sym/test.dbc)

welbouri:
Hello @ebroecker,

can I have access to implement this improvement ?

PR #732, opened 2023-11-01
Support default values for fibex signals (#731)

welbouri:
@ebroecker could you please review ?


ebroecker:
Hi @welbouri ,

thanks for your PR, 
seems to be straight forward.

I'll merge (anyway the CI is red, this seems to be another issue)

PR #734, opened 2023-11-02
Fix default values from physical to raw for fibex signals (#731)

welbouri:


welbouri:
@ebroecker, could you please review ? I was mistaken in my first PR. DEFAULT-VALUE has to be a raw value

PR #748, opened 2023-12-14
Log unsupported formats as debug instead of warning

morbult:
This commit spares the user from seeing **all** unsupported formats printed in the console when running canconvert or cancompare or just `import canmatrix` or using another package (like asammdf) which has a dependency on canmatrix. 

If a user tries to use a format that is not supported for them, an error will be logged saying that format is not supported, and that I think should be enough.


ebroecker:
should be merged already

PR #749, opened 2023-12-15
fix wrong signal id for static signals

welbouri:
Fixing the broken static signals


welbouri:
@ebroecker , could you please check and merge this ?

PR #752, opened 2023-12-22
Improve arxml Ethernet Parameter reading

xRowe:
Add arxml reading vlan, frame_size, frames_endpoints, frames.receivers/transmitters, cycle_time...etc

### DIFF
--- a/src/canmatrix/formats/arxml.py
+++ b/src/canmatrix/formats/arxml.py
@@ -1563,7 +1563,7 @@ def get_frame(frame_triggering, ea, multiplex_translation, float_factory, header
             pdu = ipdu[0]
 
         new_frame = canmatrix.Frame(ea.get_element_name(frame_elem), size=int(dlc_elem.text, 0))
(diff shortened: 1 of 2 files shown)


src/canmatrix/formats/arxml.py
```diff
@@ -1560,6 +1563,8 @@ def get_frame(frame_triggering, ea, multiplex_translation, float_factory, header
             pdu = ipdu[0]
 
         new_frame = canmatrix.Frame(ea.get_element_name(frame_elem), size=int(dlc_elem.text, 0))
+        new_frame.add_pdu(target_pdu)
```
ebroecker, line 1566: Hi @xRowe 

`target_pdu` is not defined here, this leads to an error in my tests.
Could you please have another look on it?

Thanks

xRowe:
Sorry, this line is special used in my project currently. It can be commented out.
in arxml, CAN Frame also has a PDU group

PR #754, opened 2023-12-28
Support to decode infinity value

pempem98:
Support to decode infinity value in utils


pempem98:
Hello @ebroecker,
When I use you library with my arxml, the are a tag with value equals "INF" that make the decode_number() went wrong.
Could you please review and approve it?
Thank you

PR #770, opened 2024-02-20
add a suffix containing file name in cluster name

welbouri:
The main idea is to avoid having the same default name `clusterShort` for the clusters created when converting a dbc to a fibex file.


welbouri:
@ebroecker , could you please check this MR and merge it ?

PR #771, opened 2024-02-22
fix getting file name

welbouri:
fix cases where the file name passed contains a part of the path.


welbouri:
@ebroecker , could you please check this and merge it ?

Issue #773, opened 2024-02-23
canmatrix.py


daestx:
I have an issue with doing a diff on two DBC files. I get following debug info:

 File "canmatrix\formats\__init__.py", line 79, in loadp_flat
  File "canmatrix\formats\__init__.py", line 71, in loadp
  File "canmatrix\formats\__init__.py", line 90, in load
  File "canmatrix\formats\dbc.py", line 960, in load
  File "canmatrix\canmatrix.py", line 450, in phys2raw
decimal.InvalidOperation: [<class 'decimal.DivisionUndefined'>]

coming from this line
raw_value = (self.float_factory(value) - self.float_factory(self.offset)) / self.float_factory(self.factor)

obviously, division by 0 is not handled correctly.

I changed the code as workaorund

    if self.factor == 0:
        self.factor = 1
    
    raw_value = (self.float_factory(value) - self.float_factory(self.offset)) / self.float_factory(self.factor)

please check if there is a better solution

daestx:
issue  #766 looks similar

ebroecker:
Some Idea:

canmatrix for now only uses `COMPU-INTERNAL-TO-PHYS`  compu-methods, 

Maybe your ARXML uses `COMPU-PHYS-TO-INTERNAL`.

daestx:
I used dbc file as input not an ARXML

ebroecker:
Oh sorry you are right...

Do you have a chance to check the dbc itself? 
A factor of 0 does not make sense at all. I just want to know if it's a problem with canmatrix or with the dbc file...

daestx:
So, I checked the DBC file now and identified, that the DBC is buggy. According the documentation in https://www.influxtechnology.com/post/understanding-can-dbc  

![image](https://github.com/ebroecker/canmatrix/assets/93374366/2e99c52d-1557-497c-8a35-1e99539e8d27)

So, I presume this definition of a DBC file I received is wrong, right?

![image](https://github.com/ebroecker/canmatrix/assets/93374366/38059c89-9c38-48ee-8554-728428e40fdd)

So the question to me is if such an issue shall be handled in the library

ebroecker:
Hi @daestx 

thanks for your analysis.
Yes the dbc defines the scale (=factor) to zero which is nonsense.
Seems so, that candb++ has some plausibility-check here to get around with a scale of 0.
So maybe canmatrix should do the same...

ebroecker:
should be merged

Issue #766, opened 2024-02-15
convert ARXML to DBC have error:<class 'decimal.DivisionUndefined'>


mrmuyi:
I want to convert arxml to dbc, the cmd as "canconvert -vv ess3337.arxml ess377.dbc", log as below, could you kindly help?
python version 3.8.5
canmatrix version 1.0

DEBUG - arxml - PDU: Tester_Sporadic_AtifRequestRFSDA
DEBUG - arxml - processing Frame-Trigger: Tester_BKPCANFD_AtifRequestRHRDA
DEBUG - arxml - Frame: Tester_BKPCANFD_AtifRequestRHRDA
DEBUG - arxml - PDU: Tester_Sporadic_AtifRequestRHRDA
Traceback (most recent call last):
  File "C:\Program Files\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Program Files\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Program Files\Python38\Scripts\canconvert.exe\__main__.py", line 7, in <module>
  File "C:\Program Files\Python38\lib\site-packages\click\core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "C:\Program Files\Python38\lib\site-packages\click\core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "C:\Program Files\Python38\lib\site-packages\click\core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Program Files\Python38\lib\site-packages\click\core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "C:\Program Files\Python38\lib\site-packages\canmatrix\cli\convert.py", line 155, in cli_convert
    canmatrix.convert.convert(infile, outfile, **options)
  File "C:\Program Files\Python38\lib\site-packages\canmatrix\convert.py", line 68, in convert
    dbs = canmatrix.formats.loadp(infile, **options)
  File "C:\Program Files\Python38\lib\site-packages\canmatrix\formats\__init__.py", line 71, in loadp
    return load(fileObject, import_type, key, **options)
  File "C:\Program Files\Python38\lib\site-packages\canmatrix\formats\__init__.py", line 88, in load
    dbs = module_instance.load(file_object, **options)  # type: ignore
[...]

PR #776, opened 2024-03-15
canmatrix: fix the factor could be zero

pempem98:
The factor of the signal could be zero and it makes an other error with dividing by zero so when the factor equals zero, we should set it to default value (factor=1)


pempem98:
This PR is for https://github.com/ebroecker/canmatrix/issues/773
and https://github.com/ebroecker/canmatrix/issues/766

ebroecker:
Thanks a lot!

Issue #740, opened 2023-12-01
Wrong min/max scaling when signal has slope or offset


lbeuster:
I try to convert arxml to dbc/kcd. Seems like the min/max values are scaled twice with the slope value. This happens when converting to dbc and kdc. And with all signals with a slope.

slope=0.0625
phys-max=1023.8125
internal-max=16383 (phys-max * slope)

but the converted max-value in kcd = 63.98828125 (16383 * 0.0625 * 0.0625), should be 1023.8125

In Xlsx the max value is missing.

It's the same for the offset. It is applied twice to min and max.

May be I miss something because I'm not very familiar with these CAN things?

ARXML
```
        <DATA-CONSTR UUID="626a0205-0000-0000-b658-00000004ec2a">
          <SHORT-NAME>DC_DC_IstSpannung_HV</SHORT-NAME>
          <DATA-CONSTR-RULES>
            <DATA-CONSTR-RULE>
              <PHYS-CONSTRS>
                <LOWER-LIMIT INTERVAL-TYPE="CLOSED">0.0000</LOWER-LIMIT>
                <UPPER-LIMIT INTERVAL-TYPE="CLOSED">1023.8125</UPPER-LIMIT>
                <UNIT-REF DEST="UNIT">/Unit/Unit_Unit_Volt</UNIT-REF>
              </PHYS-CONSTRS>
              <INTERNAL-CONSTRS>
                <LOWER-LIMIT INTERVAL-TYPE="CLOSED">0</LOWER-LIMIT>
                <UPPER-LIMIT INTERVAL-TYPE="CLOSED">16383</UPPER-LIMIT>
              </INTERNAL-CONSTRS>
            </DATA-CONSTR-RULE>
          </DATA-CONSTR-RULES>
        </DATA-CONSTR>

        <!-- hopefully this is the correct part ;) -->
        <COMPU-METHOD UUID="626a0202-0000-0000-b658-00000004ec2a">
          <SHORT-NAME>CM_DC_IstSpannung_HV</SHORT-NAME>
          <CATEGORY>SCALE_LINEAR_AND_TEXTTABLE</CATEGORY>
          <UNIT-REF DEST="UNIT">/Unit/Unit_Unit_Volt</UNIT-REF>
          <COMPU-INTERNAL-TO-PHYS>
            <COMPU-SCALES>
              <COMPU-SCALE>
                <LOWER-LIMIT INTERVAL-TYPE="CLOSED">0</LOWER-LIMIT>
[...]

PR #779, opened 2024-03-23
arxml: Only get min/max of the INTERNAL-CONSTR

pempem98:
Under the tag DATA-CONSTR could have PHYS-CONSTRS and INTERNAL-CONSTRS in one DATA-CONSTR-RULE and could be have more than one tag DATA-CONSTR-RULE, with current logic, we only use the LOWER-LIMIT and UPPER-LIMIT of the tag INTERNAL-CONSTRS so I modify a little to make it get the lowest internal constr value and biggest internal constr value to lower/upper variables


pempem98:
This may fix the issue https://github.com/ebroecker/canmatrix/issues/740
But still need someone do the verification

ebroecker:
Hi @pempem98,
I'll merge this one.
If you have further improvements, just create a new pr!

PR #794, opened 2024-05-28
Add Base-Data-Type and Category attributes for 32 bit length signals

SabrineBH:
Add Base-Data-Type and Category attributes for 32 bit length signals


ebroecker:
Thanks !

PR #798, opened 2024-06-11
Add Support for Github Actions workflows

xRowe:
1. Add Support for Github Actions workflows
2. When each push or PR, there will start a Action running all the test.
3. Migrate code coverage reporting from Codecov to Coveralls(Need Maintainer further support, like, ReadMe...Coveralls Permission Right...)
4. encode method modified, considering "str" value as physical value.
5. Move all the test files to root path /test/files/
6. Update Test files to satifys workflows
7. etc.....


src/canmatrix/canmatrix.py
```diff
@@ -436,25 +436,27 @@ def phys2raw(self, value=None):
...
-                )
 
-        if not (self.min <= value <= self.max):
-            logger.info(
+        try:
+            value = decimal.Decimal(value)
+        except Exception as e:
+            raise e
+
+        if not (0 <= value <= 10):
```
ebroecker, line 450: Hi @xRowe , why 0 to 10 and not self.min and self.max?

src/canmatrix/cli/compare.py
```diff
@@ -23,20 +23,22 @@
...
+
+import canmatrix.log as canmatrix_log
+
 
 logger = logging.getLogger(__name__)
 
 
 @click.command()
-@click.option('-v', '--verbose', 'verbosity', help="Output verbosity", count=True, default=1)
+@click.option('-v', '--verbose', 'verbosity', help="Output verbosity", count=True, default=False)
```
ebroecker, line 41: is this working ? 

a count-option starting with "False"?

src/canmatrix/cli/compare.py
```diff
@@ -51,23 +53,29 @@ def cli_compare(matrix1, matrix2, verbosity, silent, check_comments, check_attri
...
         verbosity = -1
-    canmatrix.log.set_log_level(root_logger, verbosity)
+    elif not verbosity:
+        # Info Message
+        verbosity = 1
+    else:
+        # Debug Message
+        verbosity = 2
+
+    canmatrix_log.set_log_level(root_logger, verbosity)
```
ebroecker, line 68: we had (theoretically) multiple verbosity levels, now we only have one? Is that what we want?

src/canmatrix/cli/compare.py
```diff
@@ -51,23 +53,29 @@ def cli_compare(matrix1, matrix2, verbosity, silent, check_comments, check_attri
...
 
     # import only after setting log level, to also disable warning messages in silent mode.
-    import canmatrix.formats  # due this import we need the import alias for log module
+    import canmatrix.formats as canmatrix_format # due this import we need the import alias for log module
```
ebroecker, line 71: what is the benefit of importing with an alias-name?

src/canmatrix/formats/arxml.py
```diff
@@ -1600,23 +1599,26 @@ def get_frame(frame_triggering, ea, multiplex_translation, float_factory, header
 
         secOC_properties = None
         if pdu is not None and 'SECURED-I-PDU' in pdu.tag:
-            payload_length = ea.get_child(pdu, "LENGTH").text
+            try:
```
ebroecker, line 1602: this would fix #799 - perfect

ebroecker (commented):
Hi @xRowe 

thank you for you work!

I have some questions added in this review. Maybe I don't understand everything just right.

Additionally:
Why don't you like the "here" in the test-scripts and use paths instead? 

Thanks again

src/canmatrix/canmatrix.py
xRowe: Test Code...
Found Issue. To be revert to normal code

src/canmatrix/cli/compare.py
xRowe: Sorry, this will be revert.
Not using count-option before

src/canmatrix/cli/compare.py
xRowe: Sorry, this will be revert too.

src/canmatrix/cli/compare.py
xRowe: Yep, Actually not usally here. During the test, I just found the import does not work, so I tried to use as alias
Change can be reverted.

xRowe:
Q: Why don't you like the "here" in the test-scripts and use paths instead?
A: Actually, I never use the PathLib before just want to use a directly way, do not know if it is the best way. Also During the test, I want to create the test file to a temp folder instead of a test folder

xRowe:
@ebroecker 
The Files should be Updated.
Could you please check again?

ebroecker:
Hi @xRowe

thank you for you work, it seems good now!

PR #802, opened 2024-06-21
Support base data type for signals with different size

SabrineBH:
Support BASE-DATA-TYPE for signals with size different than 8, 16, 32, and 64


SabrineBH:
@ebroecker Can you please check this PR and merge it ?

PR #804, opened 2024-06-21
Support base data type for diffrent bit length

SabrineBH:


SabrineBH:
@ebroecker Can you please check this PR and merge it ?

PR #817, opened 2024-10-07
fix can fd in fibex

Dalilaroussi:
@ebroecker , Regarding the CAN FD in DBC 
canmatrix does read this property from the DBC and populate the "is_fd" proeprty 
The code in canmatrix that writes back FIBEX, does not take it into consideration


ALUE:

Dalilaroussi:
@ebroecker  , could you please check this PR , thanks

PR #832, opened 2025-01-16
When converting to Fibex keep the signal instance start position as in dbc file

SabrineBH:
When converting to fibex, keep signal start position as is in dbc file

Screenshot (25)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         # os: [ubuntu-latest]
         experimental: [false]
         # python-version: ["2.7","3.4","3.5","3.6","3.7","3.8","3.9","3.10","3.11","3.12"]
-        python-version: ["3.7","3.8","3.9","3.10","3.11","3.12"]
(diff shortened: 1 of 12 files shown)


ebroecker:
Hi @SabrineBH 

your PR makes a lot of tests fail, could you have another look on it?

SabrineBH:
Unit test updated

ebroecker:
Hi @SabrineBH 

I think this PR tries to fix the issue in the wrong place.

With not looking to deep in the fibex for now, I could imagine the issue could be fixed here:

https://github.com/ebroecker/canmatrix/blob/fc2cf58a18c40e827438f334dc458d9edff75ed0/src/canmatrix/formats/fibex.py#L84

by changing from
```python
str(signal.start_bit)
```
to 

```python
str(signal.get_startbit(bit_numbering=1)
```

Maybe you can double-check my approach as I am not totally sure how the fibex should look like...

SabrineBH:
@ebroecker suggestion took into consideration, can you please check the PR. Thanks a lot!

PR #845, opened 2025-02-26
Revert "When converting to Fibex keep the signal instance start position as in dbc file"

SabrineBH:
Reverts ebroecker/canmatrix#832

### DIFF
--- a/src/canmatrix/formats/fibex.py
+++ b/src/canmatrix/formats/fibex.py
@@ -81,7 +81,7 @@ def create_signal_instance(parent, signal, signal_id):
     # type: (_Element, Signal, str) -> _Element
     signal_instance = create_sub_element_fx(parent, "SIGNAL-INSTANCE")
     signal_instance.set("ID", "PDUINST_" + signal_id)
-    create_sub_element_fx(signal_instance, "BIT-POSITION", str(signal.get_startbit(bit_numbering=1)))
+    create_sub_element_fx(signal_instance, "BIT-POSITION", str(signal.start_bit))
     create_sub_element_fx(signal_instance, "IS-HIGH-LOW-BYTE-ORDER", "false" if signal.is_little_endian else "true")
     return signal_instance
 


SabrineBH:
@ebroecker 

Apologies for creating the revert PR. The original change caused a regression, and reverting seemed like the best option at the time.

Sorry for any inconvenience. Could we proceed with merging this? Let me know your thoughts. Thanks for your understanding!

ebroecker:
Hi @SabrineBH 

what happend with the previous  patch applied? What "regression" happens with it?
Could you give me some side-info for it?

I tried to check, and I think the patch was correct, so reverting it may do it worse again.

Maybe we should also create some very small examples for testing the code regarding this issue

SabrineBH:
It's not exactly a regression in the traditional sense, but it is in the context of my solution.

Let me explain:

- For example, consider the **DBC file** ([link](https://github.com/ebroecker/canmatrix/blob/development/tests/files/dbc/test_frame_decoding.dbc)).  

  **Old behavior:**  

 The signal  
```text
  SG_ sig0 : 1|2@0+ (1,0) [0|0] ""  CCL_TEST  
```

was converted to:
```xml
<fx:SIGNAL-INSTANCE ID="PDUINST_testFrame1.sig0">
    <fx:BIT-POSITION>6</fx:BIT-POSITION>
    <fx:IS-HIGH-LOW-BYTE-ORDER>true</fx:IS-HIGH-LOW-BYTE-ORDER>
    <fx:SIGNAL-REF ID-REF="SIG_testFrame1.sig0"/>
</fx:SIGNAL-INSTANCE>
```
As you can see, the bit position changed from 1 (in the DBC) to 6 in the converted Fibex file.


  **After the fix:**  
The same signal is now converted as follows:
  ```xml
<fx:SIGNAL-INSTANCE ID="PDUINST_testFrame1.sig0">
    <fx:BIT-POSITION>1</fx:BIT-POSITION>
    <fx:IS-HIGH-LOW-BYTE-ORDER>true</fx:IS-HIGH-LOW-BYTE-ORDER>
    <fx:SIGNAL-REF ID-REF="SIG_testFrame1.sig0"/>
</fx:SIGNAL-INSTANCE>
```
This means that the bit position remains unchanged (1), which is technically correct.

However, this change impacted my solution, where I present DBC data in a tree list and order signals based on bit transmission order. Because of this, the order of signals changed, leading to incorrect encoding/decoding.

SabrineBH:
The issue that prompted this revert has been fixed through an alternative solution implemented outside this repository. 
Since the regression is no longer affecting users, this revert is no longer necessary.
Thanks for your patience @ebroecker

PR #850, opened 2025-04-15
Add support for more than 8 PDOs in EDS

willson556:
Previously, only the first 8 TPDOs and the first 8 RPDOs in an EDS would be processed. The CANopen standard supports up to 512 PDOs in each direction. With this change, all supported PDOs are processed.


ALUE:

PR #859, opened 2025-08-05
Fixes 858

### LINKED ISSUE #858: AttributeError: 'list' object has no attribute 'split' in arxml load
Some tests fail (see CI pipeline) in this method:

```
    def load(file, **options):
        # type: (typing.IO, **typing.Any) -> typing.Dict[str, canmatrix.CanMatrix]

        global frames_cache
        frames_cache = {}

        float_factory = options.get("float_factory", default_float_factory)  # type: typing.Callable
        ignore_cluster_info = options.get("arxmlIgnoreClusterInfo", False)

        decode_ethernet = options.get("decode_ethernet", False)
        decode_flexray = options.get("decode_flexray", False)

>       preferred_languages = options.get("preferred_languages", ["EN,DE"]).split(",")
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'list' object has no attribute 'split'

..\src\canmatrix\formats\arxml.py:2187: AttributeError
```

At first I thought this issue was on my branch, but I believe it to be systemic. Any thoughts?

daemonPainter:
Should fix #858, the default `preferred_language` is a list, not a string, which doesn't support `.split()`. In addition, marks `preferred_language` as optional in `Earxml.__init__()`, defaulting to `"EN,DE"`. This should make test 5 of `test_arxml.py` now green.

### DIFF
--- a/src/canmatrix/formats/arxml.py
+++ b/src/canmatrix/formats/arxml.py
@@ -53,7 +53,7 @@ _FloatFactory = typing.Callable[[typing.Any], typing.Any]
 
 
 class Earxml:
-    def __init__(self, preferred_languages):
+    def __init__(self, preferred_languages="EN,DE"):
         self.xml_element_cache = dict()  # type: typing.Dict[str, _Element]
         self.path_cache = {}
         self.sn_cache = {}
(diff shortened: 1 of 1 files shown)


ebroecker:
grazie

PR #864, opened 2025-09-02
improve generated signal for the 'updatebit'

tainnok:
!! DRAFT !!
i made big changes in my working environment and to not have just a big pull-request i tried to split it up where some mess occured ... i need to fix this. so be patient please.

signals for updatebits are initialized to 1 all the time by implizit value of class Signal initial_value.

Changes:
- explizit set of the initial_value
- add possiblity to control the inital_value of all generated updatebit-signals by the option 'update_bit_init_1' which is by default False (initial_value == False/0).
- add a comment to the generated signal


ebroecker:
should be merged with #879

PR #865, opened 2025-09-02
e2e add data_id_mode, prevent exception if PROFILE-NAME is not present

tainnok:
!! DRAFT !!
---

just know the E2E-Profile is not precise enough for some E2E-Profiles (e.g. Profile-1).

Changes:
- add the DATA-ID-MODE to the AutosarE2EProperties and fill it.
- prevent exceptions if the PROFILE-NAME is not given.
- uniform checks


ebroecker:
should be merged with #879

PR #866, opened 2025-09-02
add support for container-pdus without sub-pdu Header

tainnok:
!! DRAFT !!
---

WARNING: most likley #864 is merged before this PR.
You will get an merge-conflict at the call of get_signals() in get_frame_from_container_ipdu().
The version of this PR need to be choosen:
`        get_signals(pdu_sig_mapping, target_pdu, ea, None, float_factory, bit_offset=0,
                    generated_update_bits_init_to_1=generated_update_bits_init_to_1)`

there are two kinds of container-pdus.
- dynamic, where each sub-pdu has an header (id and length). not all sub-pdus are transmitted each time, position of sub-pdu in container-pdu is dynamic
- static, where each sub-pdu has a fixed position (offset) in the container-pdu. there is NO header for the sub-pdus!

there was some kind of support for static container-pdus, but in that case each signal of a static sub-pdu got modified start_bit (increased by the pdu-offset).
That result in a situation where the start_bit was incorrect in relation to the sub-pdu itself.
As the sub-pdu offset was not stored, the user of canmatrix has no possiblity to get the correct signal-startbit positions.

Solution: 
- store the sub-pdu offset at the sub-pdu itself
- set the startbit of the sub-pdu signals to the correct value
- add support for no-header container-pdus to the unpack() method

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -833,6 +833,8 @@ class Pdu(object):
     Whereas a PDU is the same than a frame on CAN bus, at flexray a frame may consist of
     multiple PDUs (a bit like multiple signal layout for multiplexed can frames).
     This class is only used for flexray busses.
+    Note: since container-pdus are supported for arxml, this class is also used for arxml (but only
+          for sub-pdus of container-pdus).
     """
 
     name = attr.ib(default="")  # type: str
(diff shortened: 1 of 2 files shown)


ebroecker:
should be merged with #879

PR #867, opened 2025-09-02
add support for AutosarBasePlatformTypes

tainnok:
!! DRAFT !!
---

There are (data)types which are predefined in the autosar platform.
This types are referenced in a arxml-file the same way as usaly, BUT this types are NOT part of the arxml-files.

So following the reference does NOT work.

Solution:
- add all Platform-Types
- if no types are found, try to search in the BasePlatformTypes

"Should" not have any dependecies on my other pull-requests, but i suggest to merge it AFTER #866


ebroecker:
should be merged with #879

PR #868, opened 2025-09-02
improve secoc-support

tainnok:
!! DRAFT !!
---

add support for:

- secured-i-pdus of type "cryptographic" (which use the message-link feature)
- secured-i-pdus as sub-pdus of container-pdus
- add more secoc-properties like key-id, freshness-value-id and additional informations


ebroecker:
should be merged with #879

PR #869, opened 2025-09-02
fix: do not generate invalid initial_values for Signals

tainnok:
!! DRAFT !!
---

it is possible that an arxml-file does NOT contain an initial_value for an signal.

but the class Signal need an initial_value (which make sense, because some initial-value is needed logically).

The current implementation sets the initial_value to 0 if the INITIAL_VALUE is missing in the arxml-file.
BUT if 0 is not allowed because of min/max value an inconsistent/invalid Signal is generated.

Solution: check if initial_value is between min/max. if not, set it to min.

(more or less it does not matter which value, each value between min/max is just a guess, but the current impementation make the fixed guess '0' without checking if that guess is valid)

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -455,8 +455,8 @@ class Signal(object):
         # if not (0 <= value <= 10):
         if not (self.min <= value <= self.max):
             logger.warning(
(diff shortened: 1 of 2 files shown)


ebroecker:
should be merged with #879

ebroecker:
should be merged with #879

PR #870, opened 2025-09-02
fix wrong frame-attributes because of frame-cache

tainnok:
!! DRAFT !!
---

there are real-world arxml-files where the same frame is used in several frame-triggerings.

the current implementation use some kind of frame-cache.
but canmatrix stores attributes of the frame-triggering inside the frame-object.
this, in combination with the frame-cache results in wrong generated frames.

Solution:
if we identify a frame to be already in the cache, we need to get all relevant informations from the frame_triggering and set it to the copied frame (from the cache)


ebroecker:
should be merged with #879

PR #871, opened 2025-09-02
add comment for wrong behavior, change logging-level

tainnok:
!! DRAFT !!
---

adding an comment about wrong behavior of changing the cycle-time of a frame (see comment in changes).
change the log-level


ebroecker:
should be merged with #879

PR #872, opened 2025-09-02
Reverts frame_by_id behaviour

umverteiler:
This relates to [https://github.com/ebroecker/canmatrix/pull/825](https://github.com/ebroecker/canmatrix/pull/825?utm_source=chatgpt.com).

While speeding up `frame_by_id` is useful, it changes the original behavior: Our environment expects a fresh, up-to-date frame each call, whereas the cache may return outdated data (e.g. after renaming a frame).

To stay consistent with other APIs, I’d suggest keeping two variants:
- `get_frame_by_attribute`: cache-based
- `frame_by_attribute`: always searches

Alternatively, the cache could be reset whenever a frame attribute changes, but that would require much more intrusive changes. I’d prefer keeping the PR as minimally invasive as possible.

### DIFF
--- a/src/canmatrix/canmatrix.py
+++ b/src/canmatrix/canmatrix.py
@@ -1989,7 +1989,7 @@ class CanMatrix(object):
         for element in defines_to_delete:
             del self.signal_defines[element]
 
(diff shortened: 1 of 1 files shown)


ebroecker:
should be merged with #884

PR #878, opened 2025-09-16
Support CAN-FD protocol in FIBEX if dbc file define BusType attribute as CAN FD

SabrineBH:
Support CAN-FD protocol in FIBEX if dbc file define BusType attribute as CAN FD


SabrineBH:

PR #881, opened 2025-10-15
Support baudrate custom attribute

SabrineBH:


SabrineBH:

PR #884, opened 2025-11-16
resend #872 after merging several other prs

ebroecker:


ebroecker:
this is a (identical) resent of @ChristianSteffens pr "Reverts frame_by_id behaviour #872" which I was not able to merge..

PR #885, opened 2025-11-19
Update Signal.py

SabrineBH:
Fix logger issue in Signal class


SabrineBH:

PR #890, opened 2025-12-16
add channel name based on DBName attribute

Dalilaroussi:
the aim of this MR is change the channel name based on the DBName attribute on the Dbc file , 
once we need to convert from DBC -> FIBEX ( XML)


Dalilaroussi:
@ebroecker  LGTY ?

PR #891, opened 2025-12-18
get cluster name from DBName

Dalilaroussi:
In case of DBName attribute is present , the channel and cluster could take the same name here


Dalilaroussi:
LGTY @ebroecker

